PR #26, opened 2020-11-17
Changing test and CI to run against prod

matpimenta:


GNUmakefile
```diff
@@ -11,7 +11,7 @@ PROVIDER_TYPE=rediscloud
 .PHONY: testacc install_macos website website-test
 
 testacc:
-	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m
+	TF_ACC=1 go test ./... -v $(TESTARGS) -timeout 120m -parallel=1
```
wjam, line 14: I'd probably just change the tests from `resource.ParallelTest` to `resource.Test` rather than just defining the parallel limit

internal/provider/datasource_rediscloud_data_persistence_test.go
```diff
@@ -13,8 +13,17 @@ func TestAccDataSourceRedisCloudDataPersistence(t *testing.T) {
 			{
 				Config: testAccDataSourceRedisCloudDataPersistence,
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						"data.rediscloud_data_persistence.foo", "data_persistence.#", "6"),
+					testCheckResourceValuesInListAnyOrder("data.rediscloud_data_persistence.foo", "data_persistence", []map[string]interface{}{
```
wjam, line 16: Could this be replaced with `resource.TestMatchTypeSetElemNestedAttrs` or `resource.TestCheckTypeSetElemNestedAttrs`?

internal/provider/datasource_rediscloud_subscription_test.go
```diff
@@ -53,6 +53,7 @@ data "rediscloud_payment_method" "card" {
 data "rediscloud_cloud_account" "account" {
   exclude_internal_account = true
   provider_type = "AWS" 
+  name = "PM"
```
wjam, line 56: This would force all tests to only be run on production - they should be written so that they don't assume anything about the account they are being run in

GNUmakefile
matpimenta: I think we will still try later to have parallel runs, that's why it's configurable at the test level.

internal/provider/datasource_rediscloud_data_persistence_test.go
matpimenta: 👍  I actually missed those ones, I'll try to use those

internal/provider/datasource_rediscloud_subscription_test.go
matpimenta: Yes... We actually have three options:
1. Use the cloud account 1 
2. Create cloud accounts on the fly
3. Make some environment variable that changes the `name` according to the environment.

We need make a decision on which one to go for. Thoughts?

wjam: I'd either go with 1 - as anyone wanting to contribute to the provider is going to have that cloud account available - or have a cloud account ID environment variable

trentrosenbaum (approved):

PR #43, opened 2020-11-24
Changing acc tests and preparing for production env

matpimenta:
- Making github actions call make testacc
- Combining subscription tests into a single one
- making test cloud account configurable via environment variable

### DIFF
--- a/internal/provider/provider_test.go
+++ b/internal/provider/provider_test.go
@@ -30,6 +30,10 @@ func testAccPreCheck(t *testing.T) {
 	requireEnvironmentVariables(t, RedisCloudUrlEnvVar, rediscloud_api.AccessKeyEnvVar, rediscloud_api.SecretKeyEnvVar)
 }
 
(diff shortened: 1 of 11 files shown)


internal/provider/provider_test.go
```diff
@@ -30,6 +30,10 @@ func testAccPreCheck(t *testing.T) {
 	requireEnvironmentVariables(t, RedisCloudUrlEnvVar, rediscloud_api.AccessKeyEnvVar, rediscloud_api.SecretKeyEnvVar)
 }
 
+func testAccAwsPreExistingCloudAccountPreCheck(t *testing.T) {
```
wjam, line 33: This function should be used in all of the tests that use `AWS_TEST_CLOUD_ACCOUNT_NAME` such as `TestAccDataSourceRedisCloudSubscription_basic`

wjam (approved):

internal/provider/provider_test.go
matpimenta: Good catch, I was meant to do that but forgot

PR #58, opened 2020-11-26
Dev/implement gcp vpc peering

trentrosenbaum:


internal/provider/resource_rediscloud_subscription.go
```diff
@@ -466,12 +466,20 @@ func resourceRedisCloudSubscriptionRead(ctx context.Context, d *schema.ResourceD
...
-	allowlist, err := flattenSubscriptionAllowlist(ctx, subId, api)
+	providers, err := buildCreateCloudProviders(d.Get("cloud_provider"))
 	if err != nil {
 		return diag.FromErr(err)
 	}
-	if err := d.Set("allowlist", allowlist); err != nil {
-		return diag.FromErr(err)
+
+	// CIDR whitelist is not allowed for Redis Labs internal resources subscription.
+	if redis.IntValue(providers[0].CloudAccountID) != 1 {
```
matpimenta, line 475: This will be fixed in https://github.com/RedisLabs/terraform-provider-rediscloud/pull/60
The main thing is that the length of the `providers` slice needs to be checked to work during the import.

internal/provider/resource_rediscloud_subscription_peering.go
```diff
@@ -48,35 +49,56 @@ func resourceRedisCloudSubscriptionPeering() *schema.Resource {
 				ValidateDiagFunc: validateDiagFunc(validation.StringMatch(regexp.MustCompile("^\\d+$"), "must be a number")),
 				ForceNew:         true,
 			},
+			"provider_name": {
+				Type:             schema.TypeString,
```
matpimenta, line 53: Missing a `Description` here

internal/provider/resource_rediscloud_subscription_peering_test.go
```diff
@@ -10,7 +10,9 @@ import (
 	"testing"
 )
 
-func TestAccResourceRedisCloudSubscriptionPeering_basic(t *testing.T) {
+func TestAccResourceRedisCloudSubscriptionPeering_aws(t *testing.T) {
+	t.Skip("Required environment variables currently not available under CI")
```
matpimenta, line 14: We should leave this one enabled for QA

internal/provider/resource_rediscloud_subscription_peering.go
trentrosenbaum: This has been added and matches the docs

internal/provider/resource_rediscloud_subscription_peering_test.go
trentrosenbaum: Only the GCP test is skipped now.

internal/provider/resource_rediscloud_subscription.go
trentrosenbaum: The fix from main has been merged with the length check.

matpimenta (approved):

PR #67, opened 2020-12-02
Updates to expose additional vpc peering attribues for GCP and AWS

trentrosenbaum:
This PR represents the expanding of the VPC Peering attributes to allow a GCP Peering to be create and for the terraform user to acquire enough information to activate the peering via the gcloud cli.

Both the resource and the datasource have been updated with the same schema attributes and descriptions.

Note:
- The GCP only `network_name` has been changed to `gcp_network_name`.  This brings it inline with the `gcp_project_id` as well as `redis_project_id`. and `redis_network_name`.  The aim is to make it clear for the user which values to use.
- The documentation has been updated with an example terraform output value to return the gcloud create command.
**- Need to confirm the wording on the descriptions is correct and clear.**


internal/provider/datasource_rediscloud_subscription_peerings.go
```diff
@@ -82,6 +82,26 @@ func dataSourceRedisCloudSubscriptionPeerings() *schema.Resource {
 							Type:        schema.TypeString,
 							Computed:    true,
 						},
+						"aws_peering_id": {
+							Description: "Identifier of the AWS cloud peering",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"redis_project_id": {
```
matpimenta, line 90: should we make these new fields `gcp_redis_project_id`, `gcp_redis_network_name` and `gcp_peering_id` so that everything related to GCP start with `gcp` and it's more consistent with the two other ones: `gcp_network_name` and `gcp_project_id`? 
Also `gcp_peering_id` would be more aligned to AWS's counterpart `aws_peering_id`

internal/provider/resource_rediscloud_subscription_peering.go
```diff
@@ -85,19 +85,41 @@ func resourceRedisCloudSubscriptionPeering() *schema.Resource {
 				Computed:    true,
 				ForceNew:    true,
 			},
+			"gcp_project_id": {
+				Description: "GCP project ID that the VPC to be peered lives in",
+				Type:     schema.TypeString,
```
matpimenta, line 90: the indentation looks a bit off on this file

internal/provider/datasource_rediscloud_subscription_peerings.go
```diff
@@ -165,10 +185,14 @@ func flattenVPCPeering(vpcPeerings []*subscriptions.VPCPeering) []map[string]int
 			"provider_name":  resolveProviderFromVPCPeering(currentVPCPeering),
 			"status":         currentVPCPeering.Status,
 			"aws_account_id": currentVPCPeering.AWSAccountID,
+			"aws_peering_id": currentVPCPeering.AWSPeeringID,
 			"vpc_id":         currentVPCPeering.VPCId,
 			"vpc_cidr":       currentVPCPeering.VPCCidr,
 			"gcp_project_id": currentVPCPeering.GCPProjectUID,
-			"network_name":   currentVPCPeering.NetworkName,
+			"gcp_network_name":   currentVPCPeering.NetworkName,
```
matpimenta, line 192: indentation issue here as well

internal/provider/datasource_rediscloud_subscription_peerings.go
trentrosenbaum: These have been updated to have a prefix of `gcp_redis`.  This feels clear.

matpimenta: I would still rename the `cloud_peering_id` to `gcp_peering_id`

docs/resources/rediscloud_subscription_peering.md
```diff
@@ -39,7 +42,19 @@ resource "rediscloud_subscription_peering" "example" {
...
    provider = "GCP"
    gcp_project_id = "cloud-api-123456"
-   network_name = "cloud-api-vpc-peering-example"
+   gcp_network_name = "cloud-api-vpc-peering-example"
+}
+
+output "gcloud_peering_cmd" {
+  value = <<-EOF
+  gcloud compute networks peerings create \
+  ${rediscloud_subscription_peering.example.redis_network_name} \
```
matpimenta, line 51: Maybe this one now needs to be `gcp_redis_network_name`

internal/provider/resource_rediscloud_subscription_peering_test.go
```diff
@@ -97,7 +98,10 @@ func TestAccResourceRedisCloudSubscriptionPeering_gcp(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "provider_name", "GCP"),
 					resource.TestCheckResourceAttrSet(resourceName, "status"),
 					resource.TestCheckResourceAttrSet(resourceName, "gcp_project_id"),
-					resource.TestCheckResourceAttrSet(resourceName, "network_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "gcp_network_name"),
+					resource.TestCheckResourceAttrSet(resourceName, "redis_project_id"),
```
matpimenta, line 102: I guess this will need to be changed here as well

docs/resources/rediscloud_subscription_peering.md
trentrosenbaum: Thanks for spotting this.  I made the change in my HCL working example and thought I had transferred.

internal/provider/datasource_rediscloud_subscription_peerings.go
trentrosenbaum: Renamed cloud_peering_id to gcp_peering_id.

matpimenta (approved):

PR #110, opened 2021-04-16
Updates subscription peering to not replace after import.

trentrosenbaum:
This PR focuses on correcting the provider to not try a replace against Subscription peering for the GCP cloud provider.
The peering request details returned for the GCP cloud provider allow the read to be completed and the schema for the resource is populated correctly.

If the GCP project ID is populated then the resource read function populates the rest of the GCP values and sets the provider to GCP.  The subscription ID is also set.  This should allow the import to match the HCL config.  The import test step has been added to the GCP test.

NOTE:  A Subscription peering for the AWS cloud provider will still try to replace the resource after it has been imported.  This is because the region cannot be determined at read time during the import and so this value is different.  The AWS test at this point does not have the import test step because it would lead to test failures.

### DIFF
--- a/internal/provider/resource_rediscloud_subscription_peering.go
+++ b/internal/provider/resource_rediscloud_subscription_peering.go
@@ -210,6 +210,10 @@ func resourceRedisCloudSubscriptionPeeringRead(ctx context.Context, d *schema.Re
 		return diag.FromErr(err)
 	}
 
+	if err := d.Set("subscription_id", strconv.Itoa(subId)); err != nil {
+		return diag.FromErr(err)
+	}
+
 	peerings, err := api.client.Subscription.ListVPCPeering(ctx, subId)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/provider/resource_rediscloud_subscription_peering.go
```diff
@@ -229,7 +233,15 @@ func resourceRedisCloudSubscriptionPeeringRead(ctx context.Context, d *schema.Re
 		return diag.FromErr(err)
 	}
 
-	providerName := d.Get("provider_name").(string)
+	providerName := "AWS"
+
+	if redis.StringValue(peering.GCPProjectUID) != "" {
+		providerName = "GCP"
```
bengesoff, line 239: Is there a case where the `GCPProjectUID` could get removed but it still be a GCP peering?

bengesoff (approved):

PR #129, opened 2021-07-13
Updates location of compiled provider as well as go and terraform ver…

trentrosenbaum:
This PR updates the build pipeline to build the provider with Go 1.16 and use Terraform v1.0.2.  The supportive dependencies have been updated.  Execution of the acceptance test has been adjusted to only execute an initial core set of tests.  The plugins is also being built using Terraform Plugin SDK v2.7.0

This will replace the following PR:  https://github.com/RedisLabs/terraform-provider-rediscloud/pull/127


GNUmakefile
```diff
@@ -9,6 +9,13 @@ PROVIDER_VERSION = 99.99.99
 PLUGINS_PATH = ~/.terraform.d/plugins
 PLUGINS_PROVIDER_PATH=$(PROVIDER_HOSTNAME)/$(PROVIDER_NAMESPACE)/$(PROVIDER_TYPE)/$(PROVIDER_VERSION)/$(PROVIDER_TARGET)
 
+# Use a parallelism of 4 by default for tests, overriding whatever GOMAXPROCS is
+# set to. For the acceptance tests especially, the main bottleneck affecting the
+# tests is network bandwidth and Fastly API rate limits. Therefore using the
+# system default value of GOMAXPROCS, which is usually determined by the number
+# of processors available, doesn't make the most sense.
+TEST_PARALLELISM?=3
```
bengesoff, line 17: Should this be 4 here, given the comment?

internal/provider/resource_rediscloud_subscription_test.go
```diff
@@ -14,7 +14,76 @@ import (
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func TestAccResourceRedisCloudSubscription_addUpdateDeleteDatabase(t *testing.T) {
+
+	t.Skip("Requires manual execution over CI execution")
```
bengesoff, line 85: Maybe not necessary for this PR, but might be nice to be able to configure whether the test is skipped without needing to remove this line? For example making use of `go test`'s `-short` flag to have a "CI" mode of testing:
```go
if testing.Short() {
    t.Skip("")
}
```

GNUmakefile
trentrosenbaum: When set to 4 it was erroring and using 3 gave the most consistent results against QA.

internal/provider/resource_rediscloud_subscription_test.go
trentrosenbaum: This has been implemented with the default TESTARGS supplying the `-short` flag.  The idea being if a developer wishes to execute the skipped tests then they will supply a TESTARG referencing the specific acceptance test.

GNUmakefile
```diff
@@ -9,6 +9,10 @@ PROVIDER_VERSION = 99.99.99
 PLUGINS_PATH = ~/.terraform.d/plugins
 PLUGINS_PROVIDER_PATH=$(PROVIDER_HOSTNAME)/$(PROVIDER_NAMESPACE)/$(PROVIDER_TYPE)/$(PROVIDER_VERSION)/$(PROVIDER_TARGET)
 
+# Use a parallelism of 3 by default for tests, overriding whatever GOMAXPROCS is set to.
+TEST_PARALLELISM?=3
+TESTARGS?=-short
```
bengesoff, line 14: Do you think it might be worth leaving `testacc` as the default slow mega test and having a `test-ci` target for the short mode?

trentrosenbaum: Keeping the default with -short that can be overridden when executing individual tests.

bengesoff (approved):

PR #153, opened 2021-11-05
Wrong syntax in example.

darron:
That results in this error:

```bash
Warning: Quoted references are deprecated

  on redis.tf line 47, in resource "rediscloud_subscription_peering" "dapper-eth-staging-us-west1":
  47:   provider         = "gcp"

In this context, references are expected literally rather than in quotes.
Terraform 0.11 and earlier required quotes, but quoted references are now
deprecated and will be removed in a future version of Terraform. Remove the
quotes surrounding this reference to silence this warning.


Error: Failed to query available provider packages

Could not retrieve the list of available versions for provider hashicorp/gcp:
provider registry registry.terraform.io does not have a provider named
registry.terraform.io/hashicorp/gcp
```

This fixes it.

### DIFF
--- a/docs/resources/rediscloud_subscription_peering.md
+++ b/docs/resources/rediscloud_subscription_peering.md
@@ -47,7 +47,7 @@ resource "rediscloud_subscription" "example" {
 
 resource "rediscloud_subscription_peering" "example" {
    subscription_id = rediscloud_subscription.example.id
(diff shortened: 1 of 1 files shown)


trentrosenbaum (approved):
Many thanks for catching this one and will update with the next release.

PR #207, opened 2022-05-03
Fix timing on large subscription

Hiieu:
In this PR, we've fixed the timing issue by sending a PUT request only for database blocks that were modified. That could be done by using the hashing feature for `TypeSet`. Now, the provider can detect those changes by generating a hash based on the attributes in a database block.


greg-oc (approved):
LGTM

PR #231, opened 2022-06-30
Add migration guide

Hiieu:
👋 In this PR, we've added a guide to help users with migrating their old TF config files to version`v0.0.1` of the provider.

I've uploaded the guide to a test provider in the registry so that you can check what it would look like after release.
https://registry.terraform.io/providers/Hiieu/87347/latest/docs/guides/migration-guide-v1.0.0


greg-oc (approved):
Approved with a couple of small suggestions

PR #236, opened 2022-07-11
Fix field testing issues #2

Hiieu:


docs/resources/rediscloud_subscription.md
```diff
@@ -93,16 +94,16 @@ only with Redis Labs internal cloud account
 
 The `creation_plan` block supports:
 
-* `memory_limit_in_gb` - (Required) Maximum memory usage for the initial databases
+* `memory_limit_in_gb` - (Required) Maximum memory usage that will be used your databases.
```
greg-oc, line 97:
```suggestion
* `memory_limit_in_gb` - (Required) Maximum memory usage that will be used for your largest planned database.
```

greg-oc: Would this be a better approach? As the current wording could be interpreted to mean "the maximum memory usage of the initial databases _combined_". Could do with some discussion.

docs/resources/rediscloud_subscription.md
```diff
@@ -93,16 +94,16 @@ only with Redis Labs internal cloud account
...
-* `quantity` - (Required) The planned number of databases.
-* `throughput_measurement_by` - (Required) Throughput measurement method, (either ‘number-of-shards’ or ‘operations-per-second’)
-* `throughput_measurement_value` - (Required) Throughput value (as applies to selected measurement method)
+* `replication` - (Required) Databases replication. Set to `true` if any of your databases will use replication.
+* `quantity` - (Required) The planned number of databases in the subscription.
+* `throughput_measurement_by` - (Required) Throughput measurement method that will be used by your databases, (either ‘number-of-shards’ or ‘operations-per-second’)
+* `throughput_measurement_value` - (Required) Throughput value that will be used by your databases (as applies to selected measurement method)
 * `average_item_size_in_bytes` - (Optional) Relevant only to ram-and-flash clusters. Estimated average size (measured in bytes)
-of the items stored in the database. Default: 0
+of the items stored in your databases. Default: 0
```
greg-oc, line 106: Again is this value for all of the databases combined, for the average database in the subscription, or for the largest database in the subscription?

greg-oc (commented):
I think we need to discuss the wording on a couple of fields but overall looks like good improvements

internal/provider/resource_rediscloud_subscription.go
```diff
@@ -591,13 +591,21 @@ func buildSubscriptionCreatePlanDatabases(planMap map[string]interface{}) []*sub
 	replication := planMap["replication"].(bool)
 	// Add modules to the request
 	var modules []*subscriptions.CreateModules
+	// If there is an incompatible module, then allocate it to a separate db.
+	var incompatibleModule string
```
greg-oc, line 595: I think this may be more extensible for future changes if "incompatibleModule" was made into a string list "incompatibleModules" and then these could be looped through at line 623. This would mean if other incompatible modules are added in future then they can simply be added to this list

greg-oc (approved):
Looks good, just added an optional suggestion to the incompatible module section which has been added

Hiieu:
Closing this PR, the changes are split into separate branches.

PR #239, opened 2022-07-13
Rename dummy-database to creation-plan-db

greg-oc:


Hiieu:
We should also rename those dummy / initial dbs in the doc.

Hiieu (changes requested):
We should also rename those dummy / initial dbs in the doc.

Edit: The changes in the doc are located in a separate branch.

Hiieu (approved):

PR #268, opened 2022-09-07
Fix issues with enable_tls and source_ips fields 

greg-oc:


internal/provider/resource_rediscloud_subscription_database.go
```diff
@@ -322,7 +320,8 @@ func resourceRedisCloudSubscriptionDatabaseCreate(ctx context.Context, d *schema
 
 	// Some attributes on a database are not accessible by the subscription creation API.
 	// Run the subscription update function to apply any additional changes to the databases, such as password and so on.
-	return resourceRedisCloudSubscriptionDatabaseRead(ctx, d, meta)
+	subscriptionMutex.Unlock(subId)
```
Hiieu, line 323: Isn't this equivalent of `defer subscriptionMutex.Unlock(subId)` that was deleted from the line 243? Should we leave it as it was before?

greg-oc: Not quite equivalent, because defer would release the lock after `resourceRedisCloudSubscriptionDatabaseUpdate` finishes processing, and we need the lock to be released before that line so the `Update' method can acquire the lock.  At least that seemed to be the cause of an issue I was having during testing, does that make sense?

Hiieu (approved):
LGTM

PR #293, opened 2022-12-01
Active-Active subscription resource

greg-oc:


greg-oc:
Changes depend on changes to go api client

PR #307, opened 2023-01-04
test: add `CheckDestroy` function for active-active subscriptions

bengesoff:
The active-active acceptance tests currently use the `CheckDestroy` function for `rediscloud_subscription` resources, but the tests create `rediscloud_active_active_subscription` resources instead. This always passes because the state doesn't contain the `rediscloud_subscription`s it's looking for. This PR changes the tests to use an active-active variant instead.


greg-oc (approved):
Looks good!

PR #315, opened 2023-01-11
chore: update rediscloud-go-api and add AA test sweepers

bengesoff:
Updated the rediscloud-go-api to the latest tag (on the head of the `develop` branch).

Also added sweepers for active-active subscriptions to allow cleaning up of any leaked resources during acceptance tests: https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/sweepers

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/RedisLabs/terraform-provider-rediscloud
 go 1.17
 
 require (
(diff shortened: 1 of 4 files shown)


byron-oc (approved):

bengesoff:
Two failures due to subscriptions' `state` being `error` instead of `active`. Re-running the tests as that usually seems to fix it

PR #319, opened 2023-01-19
fix: importing active-active regions failing due to incorrect ID

bengesoff:
I've added importing to the regions resource, and fixed a bug where `d.Id()` wasn't being read by the Read function, and therefore wouldn't work with the `ImportStatePassthrough` importer. Changing to read from `d.Id()` instead of `d.Get("subscription_id")` solved the issue. I also updated the test to ensure the import operation is tested


bengesoff:
Tests failing with 503s, but the changed test passes locally so should be good to go

greg-oc (approved):

PR #320, opened 2023-01-26
fix: regions inactive subscription error and memory limit hardcoding

bengesoff:
This PR fixes two issues:
1. The memory limit in the subscription creation plan was harcoded to 1000GB
2. The regions resource was producing a "subscription not active" error when modifying a few regions

I have also done some refactoring to simplify the regions resource, hopefully making the execution flows easier to follow, in order to allow me to fix the second issue. The create and update functions are now the same code, and I've tweaked some of the function signatures to make the different phases of the update more explicit in the code. The main behavioural change, other than fixing the error, is that the `delete_regions` protection flags now also apply in the create phase, as well as the update phase. However this seems to make sense to me intuitively I think? If you've disabled region deletion, to avoid accidentally deleting anything, then this should apply whether you're running Terraform for the first time, or a subsequent update in my opinion.

I have left a TODO in, for which I would love some clarification. The subscription mutex seems to be used in different ways, and I'd like to make sure it's correct before merging this. It seems to be mainly used before the "wait for subscription to be active" operation. However, in my mind, it seems like it should also encompass the actual modification operation? For example, in the `regionCreate` function, it seems like it should obtain the lock before creating the region, and release it once the subscription is active instead of only obtaining it after the region creation has succeeded.

### DIFF
--- a/.github/workflows/terraform_provider.yml
+++ b/.github/workflows/terraform_provider.yml
@@ -159,7 +159,7 @@ jobs:
           REDISCLOUD_ACCESS_KEY: ${{ secrets.REDISCLOUD_ACCESS_KEY_QA }}
           REDISCLOUD_SECRET_KEY: ${{ secrets.REDISCLOUD_SECRET_KEY_QA }}
           REDISCLOUD_URL: https://api-cloudapi.qa.redislabs.com/v1
-          AWS_TEST_CLOUD_ACCOUNT_NAME: 2024
+          AWS_TEST_CLOUD_ACCOUNT_NAME: EXTERNAL-CA-2023
 #          REDISCLOUD_ACCESS_KEY: ${{ secrets.REDISCLOUD_ACCESS_KEY_PROD }}
 #          REDISCLOUD_SECRET_KEY: ${{ secrets.REDISCLOUD_SECRET_KEY_PROD }}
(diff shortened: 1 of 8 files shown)


bengesoff:
Also depends on https://github.com/RedisLabs/rediscloud-go-api/pull/95, so the `replace` directive will need to be removed between merging that, and merging this one

pvegh-oc (approved):

bengesoff:
Failed with following 500 error:
```
2023-01-26T15:59:57.2656896Z     datasource_rediscloud_subscription_peerings_test.go:36: Step 1/1 error: Error running apply: exit status 1
2023-01-26T15:59:57.2657649Z         
2023-01-26T15:59:57.2658587Z         Error: 500 INTERNAL_SERVER_ERROR - GENERAL_ERROR: General error occurred. Please contact Redis Labs support with your task ID
2023-01-26T15:59:57.2659217Z         
2023-01-26T15:59:57.2659732Z           with rediscloud_subscription.example,
2023-01-26T15:59:57.2660521Z           on terraform_plugin_test.tf line 12, in resource "rediscloud_subscription" "example":
2023-01-26T15:59:57.2661228Z           12: resource "rediscloud_subscription" "example" {
```

Task ID 3f68eba8-6f8c-4634-a0da-a043a6f8d043

However test passed locally

PR #371, opened 2023-05-16
Add port, remote backup & CIDRs support

wjam:
Add support for:
* Specifying the port number on a normal or active/active database
* Configuring backups for normal or active/active databases
* Specifying multiple CIDR ranges when peering a normal or active/active database


wjam:
Those changes to the tests reduced the CI time from 4 hours (or 6 hours on a previous run) to 2.5 hours. While we could get a greater speed improvement by setting `parallel` back to `3`, this could potentially overload the system when the active/active databases get spun up (for example, 3 parallel tests creating active/active subscription which each have 3 databases with 3 people tests at the same time is 27 databases starting up at the same time).

greg-oc (approved):

PR #374, opened 2023-05-18
Document which parameters are ForceNew

wjam:
Add documentation for the end user to indicate which attributes will cause a resource to be replaced rather than just updated.

Also update dependencies & GitHub actions to the latest version


greg-oc (approved):
Looks good!

PR #390, opened 2023-06-19
Using a Mux Server to offer the sdkv2 provider

JohnSharpe:
[Migrated terraform test libraries](https://developer.hashicorp.com/terraform/plugin/testing/migrating) and added a [mux server](https://developer.hashicorp.com/terraform/plugin/framework/migrating/mux) to offer the existing provider.
This is in preparation for a few new resources offered by a terraform-framework provider.


provider/mux_server_test.go
```diff
@@ -0,0 +1,80 @@
...
+				muxProviderServerCreator, err := MuxProviderServerCreator(sdkProvider)
+				if err != nil {
+					return nil, err
+				}
+				return muxProviderServerCreator(), nil
+			},
+		},
+		PreCheck: func() { testAccPreCheck(t); testAccAwsPreExistingCloudAccountPreCheck(t) },
+		CheckDestroy: func(s *terraform.State) error {
+			// TODO Not happy that this method has to know what types of resource were created.
```
wjam, line 39: The solution for this might be to change all of the existing tests to use the muxed server, rather than just having a single test for the mux server?

wjam (approved):

PR #391, opened 2023-06-20
Framework provider

JohnSharpe:
Added and serving a NOP (0 resources, 0 data sources) provider built with the new/recommended terraform-framework.

Moved some resources to provider_utils.go which will be shared between both providers. The intention is for this to be absorbed into the framework-based provider when the sdk-based provider can be removed.


provider/datasource_rediscloud_cloud_account.go
```diff
@@ -46,7 +46,7 @@ func dataSourceRedisCloudCloudAccount() *schema.Resource {
 
 func dataSourceRedisCloudCloudAccountRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	var diags diag.Diagnostics
-	client := meta.(*apiClient)
+	client := meta.(*ApiClient)
```
wjam, line 49: Why does this need to be made public?

JohnSharpe: Just spotted this, can't remember why I did it. Rollback incoming...

JohnSharpe: PR looks much nicer now!

provider/debug_logger.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"log"
+	"strings"
+)
+
+type DebugLogger struct{}
+
+func (d *DebugLogger) Printf(format string, v ...interface{}) {
+	log.Printf("[DEBUG] [rediscloud-go-api] "+format, v...)
```
wjam, line 12: We should consider changing this to the new terraform-plugin-log module if it isn't much work
https://developer.hashicorp.com/terraform/plugin/log/writing

provider/framework_provider.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// Ensure RedisCloudProvider satisfies provider interface.
+var _ provider.Provider = &RedisCloudProvider{}
```
wjam, line 20: This should also probably not be public. Only make things public if you expect it to be used

provider/debug_logger.go
```diff
@@ -0,0 +1,21 @@
+package provider
+
+import (
+	"fmt"
+	"log"
+	"strings"
+)
+
+type DebugLogger struct{}
```
wjam, line 9: This shouldn't be public as well

provider/debug_logger.go
wjam: This may need changes to https://github.com/RedisLabs/rediscloud-go-api to support specifying a logger with or without a context

provider/framework_provider.go
JohnSharpe: Sure, I was thinking Java visibility - should be fixed now!

provider/mux_server.go
```diff
@@ -2,17 +2,18 @@ package provider
...
 	"context"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/providerserver"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 	"github.com/hashicorp/terraform-plugin-mux/tf5muxserver"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-func MuxProviderServerCreator(sdkProvider *schema.Provider) (func() tfprotov5.ProviderServer, error) {
+func MuxProviderServerCreator(sdkProvider *schema.Provider, frameworkProvider provider.Provider) (func() tfprotov5.ProviderServer, error) {
```
wjam, line 12: I think we might need to change this to _not_ take in the SDKv2 & framework providers as this repository is used by `https://github.com/RedisLabs/pulumi-rediscloud` and that'll need to construct the provider itself. My thinking is to make the change easy to whomever is maintaining that repository

provider/debug_logger.go
JohnSharpe: As per [this advice](https://developer.hashicorp.com/terraform/plugin/log/writing#legacy-logging), I've split into two debugLoggers, one for each provider. Pushing the change now...

provider/framework_provider.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Retrieve provider data from configuration
+	var data redisCloudProviderModel
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// All three values should be known before any other resource is applied
+	if data.Url.IsUnknown() {
```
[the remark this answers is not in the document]
wjam: I think IsUnknown returns true if the attribute wasn't defined in the provider block within the HCL, so this block - and the similar ones below - would return an error rather than use environment variables

provider/mux_server.go
JohnSharpe: That might be tricky, a lot of the tests rely on a reference to the specific provider.
Although looking at it, it seems to just be used to get an apiClient...
I'll look into it after lunch 😄

provider/framework_provider.go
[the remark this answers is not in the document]
JohnSharpe: That's not how [this documentation](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#null-and-unknown-values) reads to me, but I'll have a play around with it.

Particularly this bit:

> Because they can result from interpolations in the practitioner's config, you have no control over what attributes may contain an unknown value. However, by the time a resource is expected to be created, read, updated, or deleted, only its computed attributes can be unknown. The rest are guaranteed to have known values (or be null).

Provider configuration values can be unknown, and providers should handle that situation, even if that means just returning an error.

provider/mux_server.go
JohnSharpe: Perhaps the `MuxProviderServerCreator` method can return `(func() tfprotov5.ProviderServer, *apiClient, error)`

PR #405, opened 2023-07-12
Adding ACL User resources, data sources and documentation

JohnSharpe:


provider/datasource_rediscloud_acl_user.go
```diff
@@ -0,0 +1,92 @@
...
+				Description: "A meaningful name to identify the user",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"role": {
+				Description: "The Role which this User has.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"password": {
```
wjam, line 28: This isn't returned by the API

provider/resource_rediscloud_acl_user.go
```diff
@@ -0,0 +1,145 @@
...
+				Description: "A meaningful name to identify the user",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"role": {
+				Description: "The role which the user has",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"password": {
```
wjam, line 43: I think this'll need a ForceNew, as the value can't be read back from the API so won't be able to spot a drift

provider/datasource_rediscloud_acl_user.go
```diff
@@ -0,0 +1,86 @@
...
+		ReadContext: dataSourceRedisCloudAclUserRead,
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Description: "A meaningful name to identify the user",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"role": {
+				Description: "The Role which this User has.",
```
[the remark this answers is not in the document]
JohnSharpe: Good spot, hopefully that's fixed now (no full stops at the end)

provider/resource_rediscloud_acl_user.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	id, err := api.client.Users.Create(ctx, createUser)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(strconv.Itoa(id))
+
+	return diags
```
wjam, line 78: It's common to chain the read function up to create (& update) to ensure the correct values have been read back from the server. It's also not uncommon to call the update function at the end of the create function, especially when certain attributes can only be set _after_ a resource exists.

JohnSharpe: For the first issue (calling read after create/update), we've never seen an example where the correct values were not set. Relative to subscriptions and databases, the ACL entites are quite simple and work consistently. Would you still recommend a call to read afterwards?

As for the second issue, there are no properties that can't be set at create-time. Due to the simplicity of the resources, there's no need to update _after_ creation in these cases. Unless I'm missing the point and you're looking at something I should be updating?

wjam: I would still perform `read` at the end of `create`/`update`.

On whether `create` should call `update` - it kind of depends on the complexity of the resource being implemented. You'll see most resources in HashiCorp-maintained providers, like terraform-provider-aws, do this - but you are correct that the resource isn't sufficiently complex to need to worry about it.

wjam (approved):

PR #445, opened 2023-11-02
Added resp_version and enable_default_user to normal (flexible, not active-active) database resource and datasource

JohnSharpe:


greg-oc (approved):
Looks good!

PR #448, opened 2023-11-16
redis version on subscription and active_active subscription

JohnSharpe:


provider/resource_rediscloud_active_active_subscription.go
```diff
@@ -144,6 +144,25 @@ func resourceRedisCloudActiveActiveSubscription() *schema.Resource {
...
+					}
+
+					if old != new {
+						// The user is requesting a change
+						return false
+					}
+
+					return true
+				},
+			},
```
greg-oc, line 165:
```suggestion
			ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(regexp.MustCompile("^(default|latest)$"), "must be 'default' or 'latest'")),
			},
```
Might be good to add some validation here like we do for the cloud provider field further up. 

Also, am I correct in saying that you can't explicitly define the redis version number, it can only be defined as 'default' or 'latest'?

greg-oc (commented):
Just have one question and suggestion but otherwise looks good

provider/resource_rediscloud_active_active_subscription.go
JohnSharpe: You're right, it can only be 'default' or 'latest' but I think I wanted to let the API decide if the submission was valid because it's prone to change. Although on second thoughts I think I agree with you that we should do something here...

greg-oc (approved):
LGTM!

PR #498, opened 2024-05-14
Added a db_id property to help with flexible db lookups

JohnSharpe:


provider/datasource_rediscloud_flexible_database_test.go
```diff
@@ -74,7 +91,6 @@ resource "rediscloud_flexible_subscription" "example" {
     support_oss_cluster_api=true
     throughput_measurement_by = "operations-per-second"
     throughput_measurement_value = 1000
-	modules = ["RedisJSON", "RedisBloom"]
```
greg-oc, line 77: Is this being removed as it is unnecessary for the test and potentially slowing things down?

greg-oc (approved):

provider/datasource_rediscloud_flexible_database_test.go
JohnSharpe: That's right, it's not actually being tested anywhere and it's nice to simplify, but in practice it seems to have a negligible effect.

PR #511, opened 2024-05-23
Added PAYG properties to Essentials databases

JohnSharpe:


provider/resource_rediscloud_essentials_database.go
```diff
@@ -664,7 +762,6 @@ func resourceRedisCloudEssentialsDatabaseUpdate(ctx context.Context, d *schema.R
 		// ClientTlsCertificates
 		// Password
 		// Alerts
-		// Modules
```
greg-oc, line 667: Do we actually need any of these commented lines above?

greg-oc (approved):

provider/resource_rediscloud_essentials_database.go
JohnSharpe: No, they were more for me to say "I'm not writing them here but they will be written to the object later, with a little bit of extra logic"

provider/rediscloud_essentials_database_test.go
```diff
@@ -47,6 +47,14 @@ func TestAccRedisCloudEssentialsDatabase_BasicCRUDI(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "enable_default_user", "true"),
 					resource.TestCheckResourceAttr(resourceName, "password", "j43589rhe39f"),
 
+					resource.TestCheckResourceAttr(resourceName, "enable_payg_features", "false"),
```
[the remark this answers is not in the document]
JohnSharpe: That's right, these are just default values (because `enable_payg_features` is false)

PR #543, opened 2024-07-10
Aggregate test checks and other test refactoring

burythehammer:
- Tests now use aggregate test checks rather than simple aggressive fail checks - they should now report all checks which fail rather than the first one encountered
- Strings for resources are extracted out to variables, this is commonplace elsewhere in the tests but not in some
- Those resource strings across the whole project are now `const` as they should be immutable.


provider/rediscloud_acl_rule_test.go
```diff
@@ -18,18 +18,19 @@ func TestAccRedisCloudAclRule_DataSourceForDefaultRule(t *testing.T) {
 	testRule := "+@all -@dangerous ~*"
 	getRuleTerraform := fmt.Sprintf(getDefaultAclRuleDataSource, testName)
 
+	const AclRuleTest = "data.rediscloud_acl_rule.test"
```
JohnSharpe, line 21: V. minor, not lower case?

JohnSharpe (approved):

PR #606, opened 2025-04-10
Restructure testacc execution

trentrosenbaum:
The problem with the existing execution of the Acceptance tests is the length of time and understanding the context of failures.  Viewing the output of the failure is also problematic due to the volume of information returned.

The following PR separates out the Acceptance Tests into grouped individual steps in the workflow.  These are 

* Databases
* Subscriptions
* Subscription TLS
* Essentials Plans
* Persistence, Modules, Regions and ACLs
* Cloud Account
* Transit Payment

Each step allows the execution time to be understood and will execute 6 Acceptance Tests in parallel.
Steps that fail can be re-run if there is an unexpected situation such as network comms or timeout.


provider/resource_rediscloud_acl_rule_test.go
```diff
@@ -12,32 +12,9 @@ import (
 	"testing"
 )
 
-func TestAccRedisCloudAclRule_DataSourceForDefaultRule(t *testing.T) {
-	// This rule already exists
-	const testName = "Read-Write"
```
burythehammer, line 17: was this test being duplicated or something?

provider/resource_rediscloud_subscription_peering_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 func TestAccResourceRedisCloudSubscriptionPeering_aws(t *testing.T) {
 
+	testAccRequiresEnvVar(t, "EXECUTE_TESTS")
```
burythehammer, line 16: having now gone through a few of these, it feels like we could extract out these magic strings to some constants just incase you mistype something.

burythehammer (approved):
Couple of small things, but approved nonetheless

provider/resource_rediscloud_acl_rule_test.go
trentrosenbaum: This one was being placed in its own location file.  Some of the tests needed to be renamed and this caused a lint naming error.

provider/resource_rediscloud_subscription_peering_test.go
trentrosenbaum: Yes, agreed and I will address this in the next PR.

PR #635, opened 2025-07-31
OPCR-6: Customer Managed Key support for pro and active active subscriptions

burythehammer:
Customer Managed Key (CMK) support for Pro and AA subscriptions. This is supported through a set of new fields, and will require two `terraform apply` to go through, so the creation of the subscriptions only completes on the `update` step. This is due to a circular dependency between the RedisCloud API and GCP which would be difficult to resolve otherwise.


provider/resource_rediscloud_active_active_subscription.go
```diff
@@ -287,6 +287,42 @@ func resourceRedisCloudActiveActiveSubscription() *schema.Resource {
...
+				Default:     false,
+			},
+			"customer_managed_key_deletion_grace_period": {
+				Description: "The grace period for deleting the subscription. If not set, will default to immediate deletion grace period.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "immediate",
+			},
+			"customer_managed_key": {
+				Description: "CMK resources used to encrypt the databases in this subscription. Ignored if `customer_managed_key_enabled` set to false. Supply after the database has been put into database pending state. See documentation for CMK flow.",
```
trentrosenbaum, line 303: Confirm last part of sentence

docs/resources/rediscloud_active_active_subscription.md
```diff
@@ -19,6 +19,8 @@ subscription, then the databases defined as separate resources will be attached
 the subscription. The creation_plan block can ONLY be used for provisioning new
 subscriptions, the block will be ignored if you make any further changes or try importing the resource (e.g. `terraform import` ...).
 
+~> **Note:** The CMK (customer managed encryption key) fields require a specific flow which involves a multi step apply. Please refer to the relevant documents if using these fields.
```
trentrosenbaum, line 22: Are there any links to consider putting in here.  Maybe for a later release?

PR #765, opened 2026-02-12
OPCR-96: Framework upgrade: datasource cloud-account (rebased)

burythehammer:


provider/cloudaccount/datasource_cloud_account.go
```diff
@@ -0,0 +1,88 @@
...
+				Computed:    true,
+			},
+			"access_key_id": schema.StringAttribute{
+				Description: "The access key ID associated with the cloud account",
+				Computed:    true,
+			},
+			"exclude_internal_account": schema.BoolAttribute{
+				Description: "Whether to exclude the Redis Labs internal cloud account.",
+				Optional:    true,
+				Computed:    true,
```
burythehammer: The `Computed` flag tells terraform that the API can provide its own value for a given attribute. Since `exclude_internal_account` is a wholly provider based flag, and has no API counterpoint, this should be removed.

provider/cloudaccount/datasource_cloud_account.go
```diff
@@ -0,0 +1,88 @@
...
+
+	d.client = client
+}
+
+// Schema defines the schema for the data source.
+func (d *cloudAccountDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "The Cloud Account data source allows access to the ID of a Cloud Account configuration.  This ID can be used when creating Subscription resources.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.Int64Attribute{
```
burythehammer: right, so I remember you talking about your confusion with `id`. This is a special terraform field that worked magically under the hood in the old provider, that let it track resources in state. Now the plugin framework insists you always declare it explicitly.

But here's the catch - the old SDKv2 always defined them as `string`. So for backwards compatibility it's important we keep the attribute schema the same.

burythehammer (commented):
Looking good! Couple of small issues to fix - but they're quick adjustments.

PR #781, opened 2026-03-06
OPCR-89: Support passwordless with Block Public Endpoints

burythehammer:
## Added
- Added `enable_passwordless` attribute to `rediscloud_subscription_database` resource. When set to `true`, the database is configured without a password. Only valid when the subscription has `public_endpoint_access` disabled. Conflicts with `password`.
- Exposed `enable_passwordless` attribute on the `rediscloud_database` data source.

## Fixed
- CI: Fixed `govulncheck` step in PR workflow.


tstoykov-redis (approved):
Approved

PR #785, opened 2026-03-11
OPCR-100: Framework upgrade - datasource regions

burythehammer:
- Migrated `regions` data source from Terraform SDK v2 to the Terraform Plugin Framework. This is an internal architectural change with no breaking changes to the data source schema or behaviour.


provider/cloudaccount/datasource_cloud_account_test.go
tstoykov-redis, line 1: not a fan of mixing in changes to these files with a PR for migrating a different datasource, but I also see why

provider/regions/datasource_regions_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const regionsDataSource = "data.rediscloud_regions.example"
+
+func TestAccDataSourceRedisCloudRegions_all(t *testing.T) {
+
+	utils.AccRequiresEnvVar(t, "EXECUTE_TESTS")
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testhelpers.BasicPreCheck(t) },
```
tstoykov-redis, line 20: This isn't an exact equivalent of the previous `PreCheck` as we don't check for the `RedisCloudUrlEnvVar`. We can either add it to the `testhelpers.BasicPreCheck` function, as it is always needed, or add it as an additional check here. If you do add it to the `testhelpers.BasicPreCheck`, please also refactor the other tests in the PR, as they use it as well

tstoykov-redis (changes requested):
Only the second one I would say is a needed change

provider/regions/datasource_regions_test.go
burythehammer: Yep, i agree. let me just sort that out

tstoykov-redis (approved):

PR #788, opened 2026-03-16
OPCR-89: Passwordless feature doesn't conflict when set to false

burythehammer:
If you set passwordless to false, it would still conflict when a password is set. This fixes it so that false does not conflict.


tstoykov-redis (approved):
Fix formatting, looks good otherwise

PR #789, opened 2026-03-17
OPCR-89: Fix passwordless inconsistencies

burythehammer:
Fixing some bugs and conflicts between setting/not setting the password and passwordless fields. Now it should correctly detect a blank field (or removed) when transitioning between states.


provider/pro/rediscloud_pro_database_passwordless_test.go
```diff
@@ -80,6 +80,92 @@ func TestAccRedisCloudProDatabase_Passwordless(t *testing.T) {
...
+	password := acctest.RandString(20)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV5ProviderFactories: protoV5ProviderFactories,
+		CheckDestroy:             checkProSubscriptionDestroy,
+		Steps: []resource.TestStep{
+			// Step 1: Create with password only (no enable_passwordless field)
+			{
+				ConfigFile: config.StaticFile("testdata/pro_database_passwordless_to_password.tf"),
```
tstoykov-redis: maybe just rename the .tf file, since here the name could cause confusion

tstoykov-redis (approved):

PR #802, opened 2026-04-17
Release v2.14.1

burythehammer:
## Changed
- Migrated `regions` data source from Terraform SDK v2 to the Terraform Plugin Framework. This is an internal architectural change with no breaking changes to the data source schema or behaviour.

## Fixed
- `rediscloud_subscription` and `rediscloud_active_active_subscription`: Fixed creation plan database not being deleted after CMK subscription activation. The temporary database created during subscription provisioning is now properly cleaned up when using CMKs.
- `rediscloud_subscription_database`: Fixed inconsistent handling of `query_performance_factor` and `redis_version` when not set, ensuring nil values are not written to state as empty strings.


burythehammer:
These changes will be released as part of 2.15.0.

PR #807, opened 2026-04-23
OPCR-57: BYOC resource tags support on TF - Pro-Subscription

burythehammer:
## Added
- Added `resource_tags` attribute to the `rediscloud_subscription` resource (Pro tier). A string/string map of tags applied to the subscription's cloud infrastructure at creation time and updatable thereafter. Only supported for BYOC (Bring Your Own Cloud) subscriptions. Maximum of 30 tags; all keys and values must be lowercase.
- Exposed `resource_tags` attribute on the `rediscloud_subscription` data source.


docs/resources/rediscloud_subscription.md
```diff
@@ -95,6 +101,7 @@ The `cloud_provider` block supports:
 * `cloud_account_id` - (Optional) Cloud account identifier. Default: Redis Labs internal cloud account. **Modifying this attribute will force creation of a new resource.**
   (using Cloud Account ID = 1 implies using Redis Labs internal cloud account). Note that a GCP subscription can be created
   only with Redis Labs internal cloud account
+* `resource_tags` - (Optional) A string/string map of tags to associate with this subscription's cloud resources. All keys and values must be lowercase. Maximum of 30 tags. Only supported for BYOC (Bring Your Own Cloud) subscriptions.
```
tstoykov-redis: Just a correction, the keys and values don't have to be lowercase. Also, while it is mainly for BYOC, there is now an additional option for Redis managed cloud to be tagged by customers, so best we leave it out

provider/pro/datasource_rediscloud_pro_subscription.go
```diff
@@ -146,6 +146,12 @@ func DataSourceRedisCloudProSubscription() *schema.Resource {
 								},
 							},
 						},
+						"resource_tags": {
+							Description: "A map of resource tags associated with this subscription. Only populated for BYOC (Bring Your Own Cloud) subscriptions.",
```
tstoykov-redis: same as with the .md file, we also have some support for the Redis managed cloud to be tagged, best we leave the BYOC part out

provider/pro/resource_rediscloud_pro_subscription.go
```diff
@@ -233,6 +233,14 @@ func ResourceRedisCloudProSubscription() *schema.Resource {
 								},
 							},
 						},
+						"resource_tags": {
+							Description:      "A map of tags to associate with this subscription. All keys and values must be lowercase. Only supported for BYOC (Bring Your Own Cloud) subscriptions.",
+							Type:             schema.TypeMap,
+							Optional:         true,
+							ForceNew:         false,
+							Elem:             &schema.Schema{Type: schema.TypeString},
+							ValidateDiagFunc: validateResourceTagsFunc,
```
tstoykov-redis: Again, check for lower case may be unnecessary, it seems we may have some stale documentation, claiming only lowercase. As it's a weird case, feel free to leave it for now, and we can remove it at a later patch release

provider/pro/resource_rediscloud_pro_subscription.go
```diff
@@ -821,6 +829,23 @@ func resourceRedisCloudProSubscriptionUpdate(ctx context.Context, d *schema.Reso
 		}
```
tstoykov-redis, line 828: Should we not move the `WaitForSubscriptionToBeActive` logic in here, since we might update multiple things at a time, and get and error when we also try to send the `resource_tags`

tstoykov-redis (changes requested):

provider/pro/resource_rediscloud_pro_subscription.go
burythehammer: yeah, let's align the features between AA and pro later. I was going on the documentation. May be we don't need it at all.

tstoykov-redis (approved):

PR #808, opened 2026-04-23
OPCR-146: BYOC resource tags support on TF - Active-Active

tstoykov-redis:
rebased onto pro resource tags


provider/resource_rediscloud_active_active_subscription.go
```diff
@@ -528,6 +538,16 @@ func resourceRedisCloudActiveActiveSubscriptionRead(ctx context.Context, d *sche
 		}
 	}
 
+	if cloudDetails[0].ResourceTags != nil {
```
burythehammer: So i had a similar issue to this in my pro tags. It's a little complicated to explain, but -

Because the api schema is `omitEmpty`, a subscription with no tags becomes nil. This skips this whole logic and you'll end up keeping stale state rather than change it from something -> nothing.

Follow the logic from the pro sub and always initialise the resourcetags.

provider/resource_rediscloud_active_active_subscription.go
```diff
@@ -813,11 +857,21 @@ func buildCreateActiveActiveCloudProviders(provider string, creationPlan map[str
 		}
 	}
 
+	resourceTagsList := make([]*subscriptions.ResourceTag, 0)
+	if len(resourceTags) != 0 {
```
burythehammer: probably nitpicking, it's not that length of resourcetags is likely to return negative values, but it makes more sense to me to return this as `len(resourceTags) > 0`

docs/data-sources/rediscloud_active_active_subscription.md
```diff
@@ -47,6 +47,7 @@ output "rediscloud_active_active_subscription" {
 * `public_endpoint_access` - Whether public endpoint access is enabled for databases in the subscription
 * `maintenance_windows` - Details about the subscription's maintenance window specification, documented below
 * `pricing` - A list of pricing objects, documented below
+* `resource_tags` - A string/string map of tags to assign to the cloud resources created by this subscription.
```
burythehammer: similarly to the trivial comment on the api, I've realised we're inconsistent with trailing full stops or not on this documentation. We should probably pick one and stick with it (removing full stops probably is the more common option). We can do a later trivial PR for this across the whole documentation

provider/activeactive/active_active_subscription_resource_tags_test.go
```diff
@@ -0,0 +1,169 @@
...
+				ConfigVariables: config.Variables{
+					"rediscloud_subscription_name": config.StringVariable(subscriptionName),
+				},
+				ConfigStateChecks: []statecheck.StateCheck{
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("name"), knownvalue.StringExact(subscriptionName)),
+					statecheck.ExpectKnownValue(resourceName, tfjsonpath.New("resource_tags"), knownvalue.MapSizeExact(0)),
+					statecheck.ExpectKnownValue(datasourceName, tfjsonpath.New("resource_tags"), knownvalue.Null()),
+				},
+			},
+			// Step 4: Import subscription and verify resource tags are imported
```
burythehammer, line 91: you just cleared the tags in step 3 so this test does not actually test importing tags.

burythehammer (changes requested):
Two blocking changes and a couple of comments. Otherwise looks good.

burythehammer (approved):
You've got a tfproviderlint issue so fix that first but the code changes look good!

tstoykov-redis:
moved to https://github.com/RedisLabs/terraform-provider-rediscloud/pull/810, due to rebase

PR #813, opened 2026-04-29
OPCR-23: Remove recreate logic on AA database name change

tstoykov-redis:


burythehammer (changes requested):
You need to also wire in the name update on the Update as well. Otherwise the subscription will not update the name in place. This has worked up until now as it will recreate the DB on a name change, but there is no logic so far as I can see to update the name in place.

This will require a change on the API side I believe as the name field doesn't exist in the AA update model.

burythehammer (approved):

PR #815, opened 2026-04-29
OPCR-98: refactor: Migrate ACL Rule data source to Plugin Framework

steliyan:
## Fixed:

- Migrated `acl_rule` data source from Terraform SDK v2 to the Terraform Plugin Framework. This is an internal architectural change with no breaking changes to the data source schema or behaviour.


provider/aclrule/testdata/datasource_basic.tf
```diff
@@ -0,0 +1,7 @@
+locals {
+  name = "__NAME__"
```
burythehammer: Best if you use the new test data pattern. This involves using 

```
variable "name" {
 type = string 
}
```

here and a slightly different setup in the test (see comment there)

provider/aclrule/datasource_rediscloud_acl_rule_test.go
```diff
@@ -17,16 +17,17 @@ func TestAccDataSourceRedisCloudAclRule_ForDefaultRule(t *testing.T) {
...
-		ProtoV5ProviderFactories: protoV5ProviderFactories,
+		PreCheck:                 func() { testhelpers.BasicPreCheck(t) },
+		ProtoV5ProviderFactories: testhelpers.ProtoV5ProviderFactories(),
 		CheckDestroy:             nil, // test doesn't create a resource, so don't need to check anything
 		Steps: []resource.TestStep{
 			{
-				Config: getRuleTerraform,
+				Config: utils.RenderTestConfig(t, "./testdata/datasource_basic.tf", map[string]string{
+					"__NAME__": testName,
+				}),
```
burythehammer: to use the new config data in tests instead do the following:

```
ConfigFile: config.StaticFile("./testdata/datasource_basic.tf"),
  ConfigVariables: config.Variables{
      "name": config.StringVariable(testName),
  },
```

this is the plugin framework's recommended way of using fixture files and variables.

provider/aclrule/datasource_rediscloud_acl_rule_read.go
```diff
@@ -0,0 +1,70 @@
...
+		resp.Diagnostics.AddError(
+			"Unable to Read Cloud ACL Rules",
+			fmt.Sprintf("An error occurred while reading ACL rules: %s", err.Error()),
+		)
+		return
+	}
+
+	list = filterRules(list, filters)
+	if len(list) == 0 {
+		diags.AddError(
```
burythehammer: this `diags.AddError` will never be read by the provider as it is not appended afterwards.

`resp.diagnostics.addError`, as done previously, is fine. The errors added to diags earlier are fine too, as they are followed by a `	resp.Diagnostics.Append(diags...)`. But this diags is never added to the response afterwards, so is silently dropped.

Probably best to either stick to doing `resp.Diagnostics.AddError` every time as it means you can't forget to append them.

provider/aclrule/datasource_rediscloud_acl_rule.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+
+	d.client = client
+}
+
+func filterRules(list []*redis_rules.GetRedisRuleResponse, filters []func(*redis_rules.GetRedisRuleResponse) bool) []*redis_rules.GetRedisRuleResponse {
+	var filtered []*redis_rules.GetRedisRuleResponse
+	for _, rule := range list {
+		if filterRule(rule, filters) {
```
burythehammer: worth putting in a nilcheck here. The api probably won't return nil but it's best to be sure.

```if rule == nil { continue }```

**Review by burythehammer — CHANGES_REQUESTED** [review]
Looking great for a first contribution. A few recommended changes. The only real blocking concern I have is the swallowed diag errors.

File: provider/aclrule/datasource_rediscloud_acl_rule.go [hunk 5]
```diff
@@ -0,0 +1,90 @@
...
+			"Unexpected Data Source Configure Type",
+			fmt.Sprintf("Expected *client.ApiClient, got: %T. Please report this issue to the provider developers.", req.ProviderData),
+		)
+		return
+	}
+
+	d.client = client
+}
+
+func filterRules(list []*redis_rules.GetRedisRuleResponse, filters []func(*redis_rules.GetRedisRuleResponse) bool) []*redis_rules.GetRedisRuleResponse {
```
tstoykov-redis: Best we move the filter methods over to the `provider/aclrule/datasource_rediscloud_acl_rule_read.go` file, since that is where they are used

provider/aclrule/datasource_rediscloud_acl_rule_read.go
```diff
@@ -0,0 +1,70 @@
...
+	var data AclRuleModel
+	diags := req.Config.Get(ctx, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var filters []func(rule *redis_rules.GetRedisRuleResponse) bool
+	filters = append(filters, func(rule *redis_rules.GetRedisRuleResponse) bool {
+		return types.StringValue(redis.StringValue(rule.Name)) == data.Name
```
tstoykov-redis: Usually we compare with `redis.StringValue(rule.Name) == data.Name.ValueString()` to skip the `state` equality, but they should also be equal, so if you want to use this, best way would be `types.StringPointerValue(rule.Name) == data.Name`

provider/aclrule/datasource_rediscloud_acl_rule_read.go
```diff
@@ -0,0 +1,70 @@
...
+			"Your query returned more than one result. Please change try a more specific search criteria and try again.",
+		)
+		return
+	}
+
+	rule := list[0]
+
+	data.ID = types.StringValue(strconv.Itoa(redis.IntValue(rule.ID)))
+	data.Rule = types.StringValue(redis.StringValue(rule.ACL))
+
```
tstoykov-redis, line 79: Missing a `set` for `data.Name`. While it is fetched from the config for the filter, it's best that we still set the field from the API response, to also catch potential drift/bugs, instead of silently ignoring them. There is also the case, of the config having no filter, but still returning a valid result, which would leave the field empty

tstoykov-redis (changes requested):

tstoykov-redis (approved):

tstoykov-redis (approved):

burythehammer (approved):

PR #823, opened 2026-05-13
chore: Add a "catch-all" CI status job

steliyan:
The idea is to easily add new required steps to the `Status` job without the need to modify GitHub rulesets/branch protection rules.

### DIFF
--- a/.github/workflows/terraform_provider_pr.yml
+++ b/.github/workflows/terraform_provider_pr.yml
@@ -414,3 +414,27 @@ jobs:
         with:
           sweep_prefix: "tf-ci-pr${{ github.event.pull_request.number }}-${{ github.run_number }}"
           age_threshold: '0s'
+
+  status:
+    name: Status
+    runs-on: ubuntu-latest
+    needs:
+      - lint-complete
+      - tfproviderlint
+      - docs-validation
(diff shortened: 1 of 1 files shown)


burythehammer:
should probably target this at a 2.17.0 branch as the 2.16.0 branch is under code freeze whilst they test

.github/workflows/terraform_provider_pr.yml
```diff
@@ -414,3 +414,27 @@ jobs:
         with:
           sweep_prefix: "tf-ci-pr${{ github.event.pull_request.number }}-${{ github.run_number }}"
           age_threshold: '0s'
+
+  status:
+    name: Status
+    runs-on: ubuntu-latest
+    needs:
+      - lint-complete
```
tstoykov-redis, line 422: not blocking, but if we're depending on `lint-complete`, doesn't that already include `tfproviderlint` and `docs-validation` meaning we could remove them from the config?

tstoykov-redis (approved):

.github/workflows/terraform_provider_pr.yml
steliyan: Removed!

PR #840, opened 2026-06-02
test: Use v99.99.99 as provider version in tests

tstoykov-redis:
change to create with `99.99.99` version, instead of `dev`


steliyan (approved):
Looks good, but the PR title is a bit misleading. :)

PR #841, opened 2026-06-02
chore: Enable "unused" golangci-lint linter

steliyan:


provider/activeactive/resource_active_active_database.go
```diff
@@ -709,16 +707,6 @@ func (r *activeActiveDatabaseResource) Delete(ctx context.Context, req resource.
 	r.deleteDatabase(ctx, &state, &resp.Diagnostics)
 }
 
-// timeouts returns the resource timeouts.
-func (r *activeActiveDatabaseResource) timeouts() map[string]time.Duration {
```
tstoykov-redis, line 713: This one is the only one that worries me, but I can't seem to spot a definition within the documentation or framework that would use this, even via override

tstoykov-redis (approved):

provider/activeactive/resource_active_active_database.go
steliyan: It's private with no references. :)

That's how it was introduced: https://github.com/RedisLabs/terraform-provider-rediscloud/pull/731/changes#diff-5f9d5233ceacfae885a2fc9b041c5a1a6c03af8434b69ed492c1fef71b93dd22R647

PR #884, opened 2026-07-01
chore: migrate resource rediscloud-cloud-account

maldimirov:
Migrate the rediscloud-cloud-account resource from SDKv2 to Framework:
* Remove the hardcoded 1 minute timeout for `waitForCloudAccountToBeActive`, so that we can use the operation provided timeout instead
* Add `pending` states to the valid values while waiting for account to become active
* Update now waits for the account to return to `active` before completing
* Delete now waits for the account to actually be removed before completing

Successfully ran:

- Manual testing locally:
  - Basic CRUDI scenarios
  - Create with small timeout to force a broken state. Test that the non-activated account is persisted to state. Reconcile/untaint the resource and make it `active` in state.
- go test [TestAccResourceRedisCloudCloudAccount_basic](https://github.com/RedisLabs/terraform-provider-rediscloud/pull/884/changes#diff-21138697985aab9be44136e20836f61e5711e4ac310b690934efb72e90d8aac5)


provider/cloudaccount/resource_cloud_account.go
```diff
@@ -0,0 +1,124 @@
...
+		return
+	}
+
+	r.client = c
+}
+
+// Schema defines the schema for the resource.
+func (r *cloudAccountResource) Schema(ctx context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Creates a Cloud Account resource representing the access credentials to a cloud provider account, (`AWS` or `GCP`). Your Redis Enterprise Cloud account uses these credentials to provision databases within your infrastructure. ",
```
tstoykov-redis: NIT: Remove empty space at the end, even if present prior to migration

provider/cloudaccount/resource_cloud_account_crud.go
```diff
@@ -0,0 +1,289 @@
...
+	ctx, cancel := context.WithTimeout(ctx, readTimeout)
+	defer cancel()
+
+	id, err := strconv.Atoi(state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Invalid cloud account ID", err.Error())
+		return
+	}
+
+	notFound := r.readCloudAccountIntoModel(ctx, id, &state, &resp.Diagnostics)
```
tstoykov-redis: NIT: Make these consecutive error checks the same as in `Create` and `Update`

provider/cloudaccount/testdata/resource_basic.tf
```diff
@@ -0,0 +1,33 @@
...
+  type = string
+}
+
+resource "rediscloud_cloud_account" "test" {
+  access_key_id     = var.access_key_id
+  access_secret_key = var.access_secret_key
+  console_username  = var.console_username
+  console_password  = var.console_password
+  name              = var.name
+  provider_type     = "AWS"
```
tstoykov-redis: Would be best if we also have a variable for the `provider_type` to allow easier testing in the future

provider/cloudaccount/resource_cloud_account_test.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func TestAccResourceRedisCloudCloudAccount_basic(t *testing.T) {
+	utils.AccRequiresEnvVar(t, "EXECUTE_TESTS")
+
+	if testing.Short() {
+		t.Skip("Required environment variables currently not available under CI")
+	}
+
+	name := "tf-test-" + acctest.RandString(6)
```
tstoykov-redis: We have `TestResourcePrefix` in `utils/test_utils.go` that we can use instead of this, generates the same result

provider/cloudaccount/resource_cloud_account_test.go
```diff
@@ -0,0 +1,211 @@
...
+	if testing.Short() {
+		t.Skip("Required environment variables currently not available under CI")
+	}
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 func() { testhelpers.BasicPreCheck(t) },
+		ProtoV5ProviderFactories: testhelpers.ProtoV5ProviderFactories(),
+		Steps: []resource.TestStep{
+			{
+				ConfigFile:  config.StaticFile("./testdata/resource_invalid_provider_type.tf"),
```
tstoykov-redis: You can leave it explicitly like this, but allowing us to pass the `provider_type` means we can reuse `./testdata/resource_basic.tf` for this as well

tstoykov-redis (changes requested):

provider/cloudaccount/resource_cloud_account_test.go
```diff
@@ -0,0 +1,211 @@
...
+		"access_secret_key": config.StringVariable(os.Getenv("AWS_ACCESS_SECRET_KEY")),
+		"console_username":  config.StringVariable(os.Getenv("AWS_CONSOLE_USERNAME")),
+		"console_password":  config.StringVariable(os.Getenv("AWS_CONSOLE_PASSWORD")),
+		"name":              config.StringVariable(name),
+		"sign_in_login_url": config.StringVariable(os.Getenv("AWS_SIGNIN_URL")),
+	}
+}
+
+func TestAccResourceRedisCloudCloudAccount_basic(t *testing.T) {
+	utils.AccRequiresEnvVar(t, "EXECUTE_TESTS")
```
tstoykov-redis: Remove this check, it seems to be legacy and unnecessary when we have `TF_ACC`

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: Done

provider/cloudaccount/resource_cloud_account.go
maldimirov: Done

provider/cloudaccount/resource_cloud_account_test.go
maldimirov: Maybe even `utils.RandomWithPrefix()` since it oneshots the same `"tf-test-" + acctest.RandString(6)` that we have here,

maldimirov: Which is in no way duplicate of the ones in [sweeper_test.go](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/main/provider/sweeper_test.go#L29) :D

provider/cloudaccount/resource_cloud_account_test.go
maldimirov: As discussed offline, we are going to reuse the `resource_basic.tf` for all tests that allow it.

provider/cloudaccount/testdata/resource_basic.tf
maldimirov: As discussed offline, we are going to reuse the `resource_basic.tf` for all tests that allow it.

provider/cloudaccount/resource_cloud_account_test.go
maldimirov: Done

provider/provider_test.go
```diff
@@ -1,18 +0,0 @@
-package provider_test
-
-import (
-	"testing"
-
-	provider "github.com/RedisLabs/terraform-provider-rediscloud/provider"
-	"github.com/RedisLabs/terraform-provider-rediscloud/provider/envchecks"
-)
-
-func TestProvider(t *testing.T) {
```
tstoykov-redis, line 10: I would say we leave this test, as it is good to run a quick validation that the provider still initializes correctly. In another PR, we should probably expand it to build the muxed provider

provider/cloudaccount/resource_cloud_account_crud.go
```diff
@@ -0,0 +1,298 @@
...
+	ctx, cancel := context.WithTimeout(ctx, deleteTimeout)
+	defer cancel()
+
+	id, err := strconv.Atoi(state.ID.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Invalid cloud account ID", err.Error())
+		return
+	}
+
+	if err := r.client.Client.CloudAccount.Delete(ctx, id); err != nil {
```
tstoykov-redis, line 187: After a discussion, we can let the `NotFound` error complete in a success, but we want to log a warning, that the resource was not found, to let users know

provider/cloudaccount/resource_cloud_account_crud.go
```diff
@@ -0,0 +1,298 @@
...
+		ConsolePassword: plan.ConsolePassword.ValueStringPointer(),
+		Name:            plan.Name.ValueStringPointer(),
+		SignInLoginURL:  plan.SignInLoginURL.ValueStringPointer(),
+	})
+	if err != nil {
+		resp.Diagnostics.AddError("Failed to update cloud account", err.Error())
+		return
+	}
+
+	if _, err := waitForCloudAccountToBeActive(ctx, id, r.client, updateTimeout); err != nil {
```
tstoykov-redis, line 149: Why not also update the status that was received here, same as in Create func?

provider/cloudaccount/resource_cloud_account_crud.go
```diff
@@ -0,0 +1,298 @@
...
+			return lastStatus, lastStatus.ValueString(), nil
+		},
+	}
+	_, err := wait.WaitForStateContext(ctx)
+	return lastStatus, err
+}
+
+// waitForCloudAccountToBeDeleted waits for the cloud account to be deleted using retry.StateChangeConf.
+func waitForCloudAccountToBeDeleted(ctx context.Context, id int, c *client.ApiClient, timeout time.Duration) error {
+	deletePendingState := "pending"
```
tstoykov-redis: Lets use the `StatusPending` and `StatusDeleted` from the `cloud_accounts` model

provider/cloudaccount/resource_cloud_account_test.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// awsCloudAccountCheck verifies the AWS credentials required to create a
+// rediscloud_cloud_account resource are present.
+func awsCloudAccountCheck(t *testing.T) bool {
+	t.Helper()
+	return envchecks.RequireEnvVars(t, "AWS_ACCESS_KEY_ID", "AWS_ACCESS_SECRET_KEY", "AWS_CONSOLE_USERNAME", "AWS_CONSOLE_PASSWORD", "AWS_SIGNIN_URL")
+}
+
+func cloudAccountConfigVariables(name string) config.Variables {
```
tstoykov-redis: We should use the `ValueAndCheck` function here, for checking we have the envars and getting them

tstoykov-redis (changes requested):

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: Ok, I'll update the [relevant agents.md section](https://github.com/RedisLabs/terraform-provider-rediscloud/pull/941/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R132) as well.

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: Because, idk. I actually thought I had done this :D Will be fixed.

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: Done

provider/cloudaccount/resource_cloud_account_test.go
maldimirov: Done

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: As discussed offline, the current behavior is actually correct and TF handles the out-of-band cloud account deletion by running a "refresh state" action prior to destroy, which updates the state correctly and doesn't even try to re-delete the cloud account. So no error is thrown if going the default TF way.

provider/provider_test.go
maldimirov: Done. And a new Draft PR opened - https://github.com/RedisLabs/terraform-provider-rediscloud/pull/955

provider/cloudaccount/resource_cloud_account_crud.go
maldimirov: After checking this, the actual TF behavior is that even if we try to save something in state, when the Update operation fails, TF discards this state in favor of the previous, already saved one. The Create operation is an exception to this rule, as there is no previous state to revert to, so TF tries to save whatever it can and then mark it as "tainted".
In that context, we keep the Update behavior as is.

tstoykov-redis (approved):

tstoykov-redis (approved):

PR #892, opened 2026-07-08
RED-208027: dedupe all RequireEnvVars logic for `rediscloudApi` envs

tstoykov-redis:
This PR dedupes all methods executing the same logic as `testhelpers.BasicPreCheck` and leaves only that method


steliyan (approved):
I was thinking if we should use a different package (e.g. `tfchecks.RedisCloudAPI` instead of `testhelpers. RedisCloudAPIKeyChecks`)?

tstoykov-redis:
Might be better to do in a separate PR, as we also have a lot of duplication for similar vars for AWS and GCP as well

PR #893, opened 2026-07-08
RED-208026: rename `BasicPreCheck` to `RedisCloudCheck`

tstoykov-redis:


steliyan (approved):

steliyan (changes requested):
I think we should use `RedisCloudChecks` instead of `RedisCloudAPIKeyChecks`.

We want 3 things, not only the API key.

steliyan (approved):

PR #895, opened 2026-07-08
RED-208028: Consolidate "required environments" check

tstoykov-redis:
Dedupes all methods using the same logic as `testhelpers.RequireEnvVars`, consolidates `AWS_TEST_BYOC_CLOUD_ACCOUNT_NAME` into `AWS_TEST_CLOUD_ACCOUNT_NAME` and allows `testhelpers.RequireEnvVars` to return all missing vars at once

### DIFF
--- a/provider/activeactive/active_active_database_passwordless_test.go
+++ b/provider/activeactive/active_active_database_passwordless_test.go
@@ -23,7 +23,7 @@ func TestAccActiveActiveDatabase_Passwordless(t *testing.T) {
 	password := acctest.RandString(20)
 
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:                 func() { envchecks.RedisCloudCheck(t); testAccAwsPreExistingCloudAccountPreCheck(t) },
+		PreCheck:                 func() { envchecks.RedisCloudCheck(t); envchecks.RequireEnvVars(t, "AWS_TEST_CLOUD_ACCOUNT_NAME") },
 		ProtoV5ProviderFactories: testhelpers.ProtoV5ProviderFactories(),
 		CheckDestroy:             checkAASubscriptionDestroy,
 		Steps: []resource.TestStep{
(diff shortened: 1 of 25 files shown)


provider/pro/rediscloud_pro_subscription_datasource_resource_tags_test.go
```diff
@@ -1,29 +1,32 @@
...
 
 // TestAccDataSourceRedisCloudProSubscription_ResourceTags exercises the data
 // source's cloud_provider.0.resource_tags attribute. It requires a BYOC cloud
 // account because the API only accepts tags for BYOC subscriptions.
 func TestAccDataSourceRedisCloudProSubscription_ResourceTags(t *testing.T) {
 
-	utils.AccRequiresEnvVar(t, "EXECUTE_TESTS")
-	byocCloudAccountName := utils.AccRequiresEnvVar(t, "AWS_TEST_BYOC_CLOUD_ACCOUNT_NAME")
+	testhelpers.RequireEnvVars(t, "EXECUTE_TESTS")
+	byocCloudAccountName := os.Getenv("AWS_TEST_CLOUD_ACCOUNT_NAME")
```
steliyan: This seems wrong? It shoudl be `AWS_TEST_BYOC_CLOUD_ACCOUNT_NAME`?

provider/testhelpers/precheck.go
```diff
@@ -11,9 +12,18 @@ import (
 // variables are not set.
 func RequireEnvVars(t *testing.T, names ...string) {
 	t.Helper()
+	var missing []string
```
steliyan: I suggest we do this improvement separately. Up until now, only the first required env var was reported everywhere, right?

provider/activeactive/active_active_database_passwordless_test.go
```diff
@@ -23,7 +23,7 @@ func TestAccActiveActiveDatabase_Passwordless(t *testing.T) {
 	password := acctest.RandString(20)
 
 	resource.ParallelTest(t, resource.TestCase{
-		PreCheck:                 func() { envchecks.RedisCloudCheck(t); testAccAwsPreExistingCloudAccountPreCheck(t) },
+		PreCheck:                 func() { envchecks.RedisCloudCheck(t); envchecks.RequireEnvVars(t, "AWS_TEST_CLOUD_ACCOUNT_NAME") },
```
maldimirov, line 26: Consider moving to a separate function? E.g. testhelpers.AwsCloudAccountCheck() or similar?
I see we have the same env var check ~60 times in this PR, so maybe it makes sense to centralize it.

maldimirov (approved):

provider/activeactive/active_active_database_passwordless_test.go
maldimirov: As discussed offline, this will be a followup. Feel free to ignore this comment for this PR.

PR #899, opened 2026-07-09
tests: Make "run-testacc" a reusable workflow

steliyan:
It's pretty much copy/pasted. It makes sense to treat the Terraform acceptance tests as separate (aka reusable) workflows.

This is a good starting point to have the per-test account creation. The `pre-sweep` step would be obsolete as each run would have a brand new account and the `post-sweep` can be scoped for this specific test run.


tstoykov-redis (approved):

tstoykov-redis (approved):

.github/workflows/terraform_provider_main.yml
maldimirov, line 9: Can we remove the env vars that are now moved/redefined in `testacc.yml`? If I understand it correctly `testacc.yml` gets the values from the `secrets` input, not from these env vars here, so we don't need them anymore?

maldimirov: This is from Claude's analysis:

> Safe to drop (consumed only by acceptance tests, now covered by testacc.yml):
> AWS_TEST_CLOUD_ACCOUNT_NAME, AWS_PEERING_REGION, AWS_ACCOUNT_ID, AWS_VPC_CIDR, AWS_VPC_ID, AWS_TEST_TGW_ID, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_CONSOLE_USERNAME, AWS_CONSOLE_PASSWORD, AWS_SIGNIN_URL, GCP_VPC_PROJECT, GCP_VPC_ID, GCP_PROJECT_ID, GOOGLE_CREDENTIALS, TF_ACC, TF_LOG, and TEST_RESOURCE_PREFIX.
> 
> Note on TEST_RESOURCE_PREFIX: it looks like the sweep would need it, but the run-sweep action sets it itself at step level from its sweep_prefix input ([action.yml:18](vscode-webview://01vlkh0u4pubseni4208gva8gfdcqpn6ie833tm3cja94igu53rb/.github/actions/run-sweep/action.yml#L18)), which overrides the workflow-level value. No direct caller job reads the workflow-level one, so it's effectively dead and can go too.

steliyan: Not yet, they are needed for the pre/post-sweep.

They should be remove once we have the on-demand account creation.

steliyan: > This is from Claude's analysis:
> 
> > Safe to drop (consumed only by acceptance tests, now covered by testacc.yml):
> > AWS_TEST_CLOUD_ACCOUNT_NAME, AWS_PEERING_REGION, AWS_ACCOUNT_ID, AWS_VPC_CIDR, AWS_VPC_ID, AWS_TEST_TGW_ID, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_CONSOLE_USERNAME, AWS_CONSOLE_PASSWORD, AWS_SIGNIN_URL, GCP_VPC_PROJECT, GCP_VPC_ID, GCP_PROJECT_ID, GOOGLE_CREDENTIALS, TF_ACC, TF_LOG, and TEST_RESOURCE_PREFIX.
> > Note on TEST_RESOURCE_PREFIX: it looks like the sweep would need it, but the run-sweep action sets it itself at step level from its sweep_prefix input (action.yml:18), which overrides the workflow-level value. No direct caller job reads the workflow-level one, so it's effectively dead and can go too.

Let's keep the in-sync (way easier to reason) and they would get moved only to the `run-testacc` workflow in the end.

Yeah, it's a bit redundant, but it's error-prone to "optimize" the usage right now. :)

PR #909, opened 2026-07-13
RED-208031: Test: Expose typical env var checks for AWS and GCP

tstoykov-redis:


provider/envchecks/precheck.go
```diff
@@ -25,3 +25,35 @@ func RedisCloudCheck(t *testing.T) {
 	t.Helper()
 	RequireEnvVars(t, rediscloudApi.RedisCloudUrlEnvVar, rediscloudApi.AccessKeyEnvVar, rediscloudApi.SecretKeyEnvVar)
 }
+
+func AWSProviderCheck(t *testing.T) {
+	t.Helper()
+	RequireEnvVars(t, "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY")
+}
+
+func AWSProviderAndRegionCheck(t *testing.T) {
```
steliyan, line 34: Not a big fan of this `And` here as composition would be better.

Just noting it, maybe we will decide to change it later.

steliyan (approved):

provider/envchecks/precheck.go
tstoykov-redis: If the naming is the problem, will rename. Otherwise, planning on adding a composition function in a future PR, and will update then

PR #913, opened 2026-07-13
ci: Use Nix flakes in Makefile (2/n)

steliyan:


GNUmakefile
maldimirov, line 7: Not from this PR, but maybe it's good to have a comment that documents why this version is used.

maldimirov (approved):

tstoykov-redis (approved):

PR #914, opened 2026-07-13
chore: migrate data source Pro Subscription

maldimirov:
Migrate the Pro Subscription data source to Framework:
- Use subscription helpers from `utils` package - `FilterSubscriptions`, `FlattenMaintenance`, `FlattenPricing`, `CmkEnabledString`
- Add new helper to `utils` package - `FlattenCloudDetails`
- Fix "Self-explanatory" descriptions in docs

Successfully ran:
- Manual testing with resource + datasource
- go test [TestAccDataSourceRedisCloudProSubscription_ResourceTags](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/f8277f1e248f4e11172efaca67aa04fb46c471ef/provider/pro/rediscloud_pro_subscription_datasource_resource_tags_test.go#L17)


provider/activeactive/datasource_active_active_subscription_read.go
```diff
@@ -157,7 +157,7 @@ func (d *activeActiveSubscriptionDataSource) Read(ctx context.Context, req datas
 		)
 		return
 	}
-	pricing, diags := utils.PricingListFromAPI(ctx, pricingList)
+	pricing, diags := utils.PricingFromAPI(ctx, pricingList)
```
maldimirov, line 160: This is an AA change, but the motivation to have it in this PR is to have consistent naming with the Pro utils
Overall the motivation is that having a `List` or `Map` in the name sounds like a verb, more than a noun/type so it makes the name confusing.

provider/pro/testdata/datasource/pro_subscription.tf
```diff
@@ -48,3 +42,8 @@ resource "rediscloud_subscription_database" "example" {
   throughput_measurement_by    = "operations-per-second"
   throughput_measurement_value = 1000
 }
+
+data "rediscloud_subscription" "example" {
```
tstoykov-redis, line 46: Didn't we want to separate the datasource from the subscription .tf files?

provider/pro/datasource_pro_subscription_test.go
```diff
@@ -0,0 +1,110 @@
...
+			{
+				ConfigFile: config.StaticFile(proSubscriptionDataSourceConfigPath),
+				ConfigVariables: config.Variables{
+					"subscription_name": config.StringVariable(name),
+				},
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(dataSourceName, "name", name),
+					resource.TestCheckResourceAttr(dataSourceName, "payment_method", "credit-card"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "payment_method_id"),
+					resource.TestMatchResourceAttr(dataSourceName, "memory_storage", regexp.MustCompile("ram")),
```
tstoykov-redis, line 41: I know this is just copy pasted, but is there any specific reason to check like this, and nto via the `resource.TestCheckResourceAttr` method?

provider/pro/datasource_pro_subscription_test.go
```diff
@@ -0,0 +1,110 @@
...
+			checkAASubscriptionDestroy,
+		),
+		Steps: []resource.TestStep{
+			{
+				ConfigFile: config.StaticFile(subscriptionFiltersConfigPath),
+				ConfigVariables: config.Variables{
+					"subscription_name": config.StringVariable(name),
+				},
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttrPair(proDataSourceName, "id", proResourceName, "id"),
```
tstoykov-redis, line 104: We also need to add a check, that ensures we didn't leak the Pro results in AA and vise versa. That was what the old test verified

provider/pro/helpers_test.go
```diff
@@ -14,13 +14,21 @@ import (
 // checkProSubscriptionDestroy verifies that all rediscloud_subscription resources
 // have been destroyed after a test completes.
 func checkProSubscriptionDestroy(s *terraform.State) error {
```
tstoykov-redis, line 16: Export these functions so they can be reused in other checks as well. This logic is duplicated multiple times in the project

provider/utils/subscription.go
```diff
@@ -130,3 +140,118 @@ func ResourceTagsFromAPI(ctx context.Context, tags []*subscriptions.ResourceTag)
...
+
+	return types.SetValueFrom(ctx, regionType, models)
+}
+
+func cloudNetworksFromAPI(ctx context.Context, networks []*subscriptions.Networking) (types.List, diag.Diagnostics) {
+	networkType := types.ObjectType{AttrTypes: cloudNetworkAttrTypes}
+	models := make([]cloudNetworkModel, 0, len(networks))
+
+	for _, network := range networks {
+		models = append(models, cloudNetworkModel{
```
tstoykov-redis, line 249: Add a TODO 3.0 comment to replace this with the direct `types.StringPointerValue`

provider/activeactive/datasource_active_active_subscription_read.go
tstoykov-redis: This is okay as a change

provider/utils/subscription.go
```diff
@@ -130,3 +140,118 @@ func ResourceTagsFromAPI(ctx context.Context, tags []*subscriptions.ResourceTag)
 	}
 	return types.MapValueFrom(ctx, types.StringType, result)
 }
+
```
tstoykov-redis, line 143: I think it would be best to move all the structs here for models to a shared subscription_models file, for easier viewing and consistent separation. Including the PricingModel that came from AA changes

provider/utils/subscription.go
```diff
@@ -130,3 +140,118 @@ func ResourceTagsFromAPI(ctx context.Context, tags []*subscriptions.ResourceTag)
...
+	regionType := types.ObjectType{AttrTypes: cloudRegionAttrTypes}
+	models := make([]cloudRegionModel, 0, len(regions))
+
+	for _, region := range regions {
+		networks, diags := cloudNetworksFromAPI(ctx, region.Networking)
+		if diags.HasError() {
+			return types.SetNull(regionType), diags
+		}
+
+		availabilityZones := make([]string, 0, len(region.PreferredAvailabilityZones))
```
tstoykov-redis, line 222: we can replace this list creation and iteration with `redis.StringSliceValue` which implements the same logic

provider/utils/subscription.go
```diff
@@ -130,3 +140,118 @@ func ResourceTagsFromAPI(ctx context.Context, tags []*subscriptions.ResourceTag)
...
+
+		resourceTags, diags := ResourceTagsFromAPI(ctx, cloudDetail.ResourceTags)
+		if diags.HasError() {
+			return types.ListNull(cloudProviderType), diags
+		}
+
+		models = append(models, cloudProviderModel{
+			Provider:       types.StringValue(redis.StringValue(cloudDetail.Provider)),
+			CloudAccountID: types.StringValue(strconv.Itoa(redis.IntValue(cloudDetail.CloudAccountID))),
+			AWSAccountID:   types.StringValue(redis.StringValue(cloudDetail.AWSAccountID)),
```
tstoykov-redis, line 203: Specifically `AWSAccountID` we can use the `types.StringPointerValue` directly, as it was only set if not null

provider/utils/subscription_test.go
```diff
@@ -166,3 +178,143 @@ func TestResourceTagsMapFromAPI(t *testing.T) {
 		assert.Empty(t, m.Elements())
 	})
 }
+
+type testNetwork struct {
```
tstoykov-redis, line 182: If you export the types from `subscription.go` as mentioned in that comment, you would be able to reuse them here, making this cleaner, as was already done with PricingModel

provider/utils/subscription_test.go
```diff
@@ -166,3 +178,143 @@ func TestResourceTagsMapFromAPI(t *testing.T) {
...
+		list, diags := utils.CloudProvidersFromAPI(ctx, []*subscriptions.CloudDetail{cd})
+		require.False(t, diags.HasError())
+
+		var cps []testCloudProvider
+		require.False(t, list.ElementsAs(ctx, &cps, false).HasError())
+		require.Len(t, cps, 1)
+
+		assert.Equal(t, "GCP", cps[0].Provider.ValueString())
+		assert.Equal(t, "0", cps[0].CloudAccountID.ValueString())
+		assert.False(t, cps[0].AwsAccountID.IsNull())
```
tstoykov-redis, line 283: This would also need to assert `True` with the change for `AwsAccountID`

provider/pro/datasource_pro_subscription_read.go
```diff
@@ -0,0 +1,145 @@
...
+
+	model.ID = types.StringValue(strconv.Itoa(subId))
+	model.Name = types.StringValue(redis.StringValue(sub.Name))
+
+	paymentMethodID := ""
+	if sub.PaymentMethodID != nil {
+		paymentMethodID = strconv.Itoa(redis.IntValue(sub.PaymentMethodID))
+	}
+	model.PaymentMethodID = types.StringValue(paymentMethodID)
+	model.PaymentMethod = types.StringValue(redis.StringValue(sub.PaymentMethod))
```
tstoykov-redis, line 78: This can just use the `types.StringPointerValue` as it didn't use the `redis.StringValue` method before

tstoykov-redis (changes requested):

provider/pro/testdata/datasource/pro_subscription.tf
maldimirov: Addressed in [5f3add4](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/5f3add426bcb3bbdf54add10ebb7a660ac37851a). The data source now has its own fixture file and the test loads the shared fixture directory.

provider/pro/datasource_pro_subscription_test.go
maldimirov: Addressed in [bba9dee](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/bba9deef844ea9f4cc0be0fa8c7b84d38df4ada2). The test now checks the exact `ram` value with `resource.TestCheckResourceAttr`.

provider/pro/datasource_pro_subscription_test.go
maldimirov: Addressed in [4b65adf](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/4b65adf4038f071c59e44adbb27a7b1ef00982ec). The unit test now filters same-name Pro and Active-Active subscriptions independently, asserts that each result contains exactly one subscription, and verifies the returned subscription identity.

provider/pro/helpers_test.go
maldimirov: Addressed in [ae84774](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/ae84774d0c2addfb838bbb88c4b84df03e0c360b). The destroy checks now live as exported helpers in `provider/testhelpers`, and the existing callers use those shared implementations.

provider/utils/subscription.go
maldimirov: Addressed in [93ef7eb](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/93ef7eba78ad94b008ebb04c05a46189eeebc1e9). The network conversion now carries the Terraform 3.0 TODO for migrating to `types.StringPointerValue`.

provider/utils/subscription.go
maldimirov: Addressed in [72b5d7f](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/72b5d7fc083f2f87c43e84e455894c3283bbea94). Pricing and cloud models now live together in the shared `subscription_models.go` file.

provider/utils/subscription.go
maldimirov: Addressed in [31405e4](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/31405e4765bfe50808504a3969534b02a398ac35). Preferred availability zones now use `redis.StringSliceValue` while preserving a known empty Terraform list for an absent API collection.

provider/utils/subscription.go
maldimirov: Addressed in [61bdf77](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/61bdf7728188868b0d8e9255a8b088f6880357cc). `AWSAccountID` now uses `types.StringPointerValue`, preserving null when the API value is absent.

provider/utils/subscription_test.go
maldimirov: Addressed in [250be66](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/250be668017f3cc3f0f7345af404c80088c46cc7). The mapping tests now decode into the exported production cloud models instead of duplicating test-only model structs.

provider/utils/subscription_test.go
maldimirov: Addressed in [61bdf77](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/61bdf7728188868b0d8e9255a8b088f6880357cc). The nil AWS account case now explicitly asserts that the Terraform value is null.

provider/pro/datasource_pro_subscription_read.go
maldimirov: Discussed offline and agreed this is not relevant

tstoykov-redis: This is an incorrect assumption, ignore it

provider/pro/datasource_pro_subscription_test.go
tstoykov-redis: I meant an assertion within this test as well. We can just add a check, that verifies the AA and Pro subscriptions are not the same via an ID check

tstoykov-redis (changes requested):

provider/pro/datasource_pro_subscription_test.go
maldimirov: Addressed in [b8158a5](https://github.com/RedisLabs/terraform-provider-rediscloud/commit/b8158a57b9d1859c17efb6ed721747dc1f1b0944). The acceptance test now explicitly compares the Pro and Active-Active data source IDs and requires them to differ, documenting why this guards against cross-deployment leakage.

PR #916, opened 2026-07-14
RED-208675: Test: Dedupe SharedTestClient

tstoykov-redis:


provider/activeactive/helpers_test.go
```diff
@@ -43,7 +29,7 @@ func testRandomWithPrefix(n ...int) string {
 // resources have been destroyed. Uses terraform-plugin-testing's terraform.State
 // (required for ConfigFile/ConfigVariables gold standard test pattern).
 func checkAASubscriptionDestroy(s *terraform.State) error {
-	apiClient, err := getTestClient()
+	testApiClient, err := client.GetTestClient()
```
steliyan: Let's keep the `apiClient` convention. Having `Test` in the factory method name itself is enough.

steliyan (approved):

provider/client/client.go
```diff
@@ -29,3 +37,23 @@ func NewClient() (*ApiClient, error) {
 		Client: client,
 	}, nil
 }
+
+func SharedTestClient(t *testing.T) *ApiClient {
+	sharedClientOnce.Do(func() {
```
maldimirov: Consider calling the `GetTestClient()` here, as both functions touch/use the same `sharedClient` and `sharedClientErr` variables. Yes, they are both guarded by the same `sharedClientOnce`, but still I think in this situation it's better to DRY out the code :)

provider/client/client.go
```diff
@@ -29,3 +37,23 @@ func NewClient() (*ApiClient, error) {
 		Client: client,
 	}, nil
 }
+
+func SharedTestClient(t *testing.T) *ApiClient {
```
maldimirov: Nit: Lets add a short comment/description

maldimirov: Nit: Consider `GetSharedTestClient()` just for the sake of keeping with the `GetXXX` naming pattern/convention.

steliyan: I would say better to use `MustXXX` (e.g. https://dev.to/dubjay18/gos-must-pattern-streamline-your-error-handling-27ff).

As this just unpacks the error.

provider/utils/test_utils.go
```diff
@@ -4,47 +4,21 @@ import (
...
 // CheckNoDatabasesForSubscription verifies that no databases exist in a subscription.
 // Combines GetTestClient with CheckNoDatabasesInSubscription for convenience in test check functions.
 func CheckNoDatabasesForSubscription(ctx context.Context, subId int) error {
-	api, err := GetTestClient()
+	apiClient, err := client.GetTestClient()
 	if err != nil {
 		return err
 	}
-	return CheckNoDatabasesInSubscription(ctx, subId, api)
+	return CheckNoDatabasesInSubscription(ctx, subId, apiClient)
```
maldimirov, line 21: Separate PR/Discussion: It seems `CheckNoDatabasesInSubscription()` is only used by `CheckNoDatabasesForSubscription()`, so it doesn't make sense for these to be separate functions. I'd say we remove `CheckNoDatabasesForSubscription()` and keep only `CheckNoDatabasesForSubscription()`, which will create/get the client itself.

provider/rediscloud_active_active_subscription_test.go
```diff
@@ -115,8 +116,8 @@ func TestAccResourceRedisCloudActiveActiveSubscription_CRUDI_Redis7(t *testing.T
 							return err
 						}
 
-						client := sharedTestClient(t)
-						sub, err := client.Client.Subscription.Get(context.TODO(), subId)
+						apiClient := client.SharedTestClient(t)
+						sub, err := apiClient.Client.Subscription.Get(context.TODO(), subId)
```
maldimirov, line 120: Total NIT (discussion): Something feels very wrong with the tautology in `apiClient.Client`.
I'm thinking, what if we call this var `capi`. And then:
- We use it like `capi.Client.Subscription.Get()`
- Respectively the functions that create it will be:
  -  `client.GetCapi()` - for `CheckDestroy`, etc.
  - And `client.GetSharedCapi()` or `client.GetCapiForTests()` or `client.GetTestingCapi()` for test functions

maldimirov (approved):

provider/utils/test_utils.go
tstoykov-redis: Agreed, but yes, different PR

provider/rediscloud_active_active_subscription_test.go
tstoykov-redis: I don't like the idea of using `capi`, as that could also not be correct in the future. We could go with just `api` so `api.Client.Subscription.Get()`.
We can change the return to directly return a `rediscloudApi.Client` client instead, but I think that would be for a different PR, as this type of logic is everywhere in the project

PR #922, opened 2026-07-17
chore: migrate data source Active Active Subscription

maldimirov:
Migrate the data source Active Active Subscription from SDKv2 to Framework (using essentials-plan as blueprint):
- Shared helpers go to provider/utils/subscription.go. The follow-up pro-subscription migration will reuse them and drop the old copies
- maintenance_windows/pricing migrated to blocks
- Add sorting in FlattenPricing - Pricing.List returns regions in nondeterministic order, which results in a potential perpetual tf plan diff
- Use variables from the `rediscloud-go-api`, instead of hardcoded strings like "active-active"
- Use named filter function types
- Add unit tests for standalone/helper functions
- Drop the `pro.CMK_ENABLED_STRING` in favor of `utils.CmkEnabledString`, that lives in the shared `utils/subscription` package - better place and better name :)
- Separate the ACC tests for the resource and data source and add a standalone data source ACC test

Successfully ran:

- Manual testing with resource + datasource
- go test [TestAccResourceRedisCloudActiveActiveSubscription_CRUDI_Redis7](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/f8277f1e248f4e11172efaca67aa04fb46c471ef/provider/rediscloud_active_active_subscription_test.go#L26)
- go test [TestAccResourceRedisCloudActiveActiveSubscription_CRUDI_Redis8](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/f8277f1e248f4e11172efaca67aa04fb46c471ef/provider/rediscloud_active_active_subscription_test.go#L278)
- go test [All ResourceTags tests](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/f8277f1e248f4e11172efaca67aa04fb46c471ef/provider/activeactive/active_active_subscription_resource_tags_test.go)


provider/activeactive/datasource_active_active_subscription.go
```diff
@@ -0,0 +1,199 @@
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "The identifier of the subscription",
+				Computed:    true,
+			},
+			"name": schema.StringAttribute{
+				Description: "A meaningful name to identify the subscription",
+				Required:    true,
+			},
+			"payment_method": schema.StringAttribute{
```
tstoykov-redis, line 67: This should only be the first sentence, as this is just a datasource, and we don't pass this value at all. Also, same follows for the documentation. It should be cleaned up, as it seems the docs from the resource were used there, which is not correct

docs/data-sources/rediscloud_active_active_subscription.md
tstoykov-redis: Not changed now, but incorrect, since the datasource doesn't set `payment_method`. A review of the others is also probably needed, since it seems we reused the docs from the resource

provider/datasource_rediscloud_active_active_subscription_regions.go
```diff
@@ -100,7 +100,7 @@ func dataSourceRedisCloudActiveActiveRegionsRead(ctx context.Context, d *schema.
 
 	// Filter to active-active subscriptions only (pro subs come from the same endpoint)
 	filters = append(filters, func(sub *subscriptions.Subscription) bool {
```
tstoykov-redis, line 102: for a separate PR, but we should replace this with the newly created ActiveActiveSubscriptionFilter to keep consistency

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
 
+## 2.19.0 (TBD)
+
+### Changed
+
+- Migrated `rediscloud_active_active_subscription` data source from Terraform SDK v2 to the Terraform Plugin Framework. This is an internal architectural change with no schema changes; the only behaviour change is the `null`-vs-empty-string handling noted under Fixed.
+
+### Fixed
+
+- `rediscloud_active_active_subscription` data source: optional attributes the API omits (`payment_method_id`, `customer_managed_key_deletion_grace_period`, `customer_managed_key_redis_service_account`, `customer_managed_key_aws_role_arn`) are now written to state as `null` instead of empty strings.
```
tstoykov-redis: We shouldn't change this here, as it is a breaking change, best to revert this behavior and the test checks associated with it

provider/utils/subscription.go
```diff
@@ -0,0 +1,206 @@
...
+
+// SubscriptionNameFilter matches subscriptions with the given name.
+func SubscriptionNameFilter(name string) SubscriptionFilter {
+	return func(sub *subscriptions.Subscription) bool {
+		return redis.StringValue(sub.Name) == name
+	}
+}
+
+// maintenanceWindowAttrTypes describes a single window within a maintenance_windows block.
+var maintenanceWindowAttrTypes = map[string]attr.Type{
```
tstoykov-redis: Lets convert these *AttrTypes and the corresponding flatten methods into Framework Custom Types, as explained [here](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom). Make sure to implement the methods listed there, to cover basic functionality

provider/activeactive/datasource_active_active_subscription_read.go
```diff
@@ -0,0 +1,155 @@
...
+		return
+	}
+
+	sub := subs[0]
+	subId := redis.IntValue(sub.ID)
+
+	model.ID = types.StringValue(strconv.Itoa(subId))
+	model.Name = types.StringPointerValue(sub.Name)
+
+	if sub.PaymentMethodID != nil {
```
tstoykov-redis, line 80: As mentioned in the Changelog comment, we don't want to change the return behavior so far, since we returned "" till now, this might be a breaking change

provider/activeactive/datasource_active_active_subscription_read.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	model.PaymentMethod = types.StringPointerValue(sub.PaymentMethod)
+	model.NumberOfDatabases = types.Int64Value(int64(redis.IntValue(sub.NumberOfDatabases)))
+	model.Status = types.StringPointerValue(sub.Status)
+
+	cmkEnabled := sub.PersistentStorageEncryptionType != nil &&
+		redis.StringValue(sub.PersistentStorageEncryptionType) == utils.CmkEnabledString
+	model.CustomerManagedKeyEnabled = types.BoolValue(cmkEnabled)
+
+	if sub.CustomerManagedKeyAccessDetails != nil {
```
tstoykov-redis, line 95: again the same here, this would be a breaking change

provider/activeactive/datasource_active_active_subscription_read.go
```diff
@@ -0,0 +1,155 @@
...
+		model.CustomerManagedKeyRedisServiceAccount = types.StringPointerValue(sub.CustomerManagedKeyAccessDetails.RedisServiceAccount)
+		model.CustomerManagedKeyAwsRoleArn = types.StringPointerValue(sub.CustomerManagedKeyAccessDetails.AwsRoleArn)
+	} else {
+		model.CustomerManagedKeyRedisServiceAccount = types.StringNull()
+		model.CustomerManagedKeyAwsRoleArn = types.StringNull()
+	}
+
+	model.CustomerManagedKeyDeletionGracePeriod = types.StringPointerValue(sub.DeletionGracePeriod)
```
tstoykov-redis: Same here, breaking

tstoykov-redis (changes requested):

provider/activeactive/datasource_active_active_subscription.go
maldimirov: Good catch — trimmed `payment_method` to just the descriptive first sentence (dropped the create-time resource wording), and cleaned up the other descriptions that had leaked in from the resource. Done in a5e9ca8f97db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/a5e9ca8f97db

docs/data-sources/rediscloud_active_active_subscription.md
maldimirov: Reviewed and cleaned up the data-source docs: `payment_method` no longer carries the resource-only wording (`(Optional)`, default, ignored-after-creation), dropped the imperative sentence on `customer_managed_key_aws_role_arn`, and fixed the `pricing` attribute names that were still camelCase (copied from the resource docs) to the actual snake_case names. Done in a5e9ca8f97db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/a5e9ca8f97db

CHANGELOG.md
maldimirov: Agreed, reverted — the data source keeps returning `""` (not `null`) for these fields to stay SDKv2-compatible. Removed the CHANGELOG entry describing the null change and updated the associated test assertions. Done in 240dbe6950db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/240dbe6950db

provider/activeactive/datasource_active_active_subscription_read.go
maldimirov: Reverted to the SDKv2 `""` behavior here to avoid the breaking change. Done in 240dbe6950db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/240dbe6950db

provider/activeactive/datasource_active_active_subscription_read.go
maldimirov: Same — reverted to `""` here too. Done in 240dbe6950db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/240dbe6950db

provider/activeactive/datasource_active_active_subscription_read.go
maldimirov: Same — reverted to `""` here too. Done in 240dbe6950db: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/240dbe6950db

provider/utils/subscription.go
maldimirov: Done — converted `pricing` into a Framework custom type as the reference implementation: a new `provider/customtypes` package with `PricingType`/`PricingValue` implementing the interface methods from the doc (`Equal`, `String`, `ValueFromObject`, `ValueFromTerraform`, `ValueType`, `Type`), built through a `tfsdk`-tagged `PricingModel` (compile-checked, no string-keyed maps) with the attr-types derived from it. The same pattern applies to the maintenance AttrTypes — happy to convert those here or in a follow-up, whichever you prefer. Done in 19eb548cbe90: https://github.com/RedisLabs/terraform-provider-rediscloud/commit/19eb548cbe90

provider/datasource_rediscloud_active_active_subscription_regions.go
maldimirov: Agreed, and as you noted this is best as a separate PR. `utils.ActiveActiveSubscriptionFilter()` is added in this PR, so the SDKv2 AA-regions data source can be repointed to it in a follow-up — tracking it separately.

maldimirov: Follow-up: reverted the incidental change here so this PR no longer touches the file — https://github.com/RedisLabs/terraform-provider-rediscloud/commit/63fd10813ab232b1ffaa149ffc705eaa92af0d47

The full repoint onto the shared `utils` filters is here - https://github.com/RedisLabs/terraform-provider-rediscloud/pull/981

docs/data-sources/rediscloud_active_active_subscription.md
```diff
@@ -34,21 +34,19 @@ output "rediscloud_active_active_subscription" {
...
-  Changes to) this attribute are ignored after creation.**
+* `payment_method` - The payment method for the subscription, (either `credit-card` or `marketplace`).
 * `payment_method_id` - A valid payment method pre-defined in the current account
 * `cloud_provider` - The cloud provider used with the subscription, (either `AWS` or `GCP`).
 * `number_of_databases` - The number of databases that are linked to this subscription.
 * `status` - Current status of the subscription
 * `customer_managed_key_enabled` - Whether customer managed key encryption is enabled for the subscription
 * `customer_managed_key_deletion_grace_period` - The deletion grace period for the customer managed key (e.g. 'immediate', '15-minutes')
 * `customer_managed_key_redis_service_account` - The Redis service account principal associated with the subscription. This is used to grant access to the customer managed encryption key (GCP subscriptions).
-* `customer_managed_key_aws_role_arn` - The ARN of the IAM role used by the subscription to access the AWS KMS customer managed key. Grant this role access to your KMS key via key policy (AWS subscriptions).
```
tstoykov-redis, line 47: I think we can leave this as it was, might be useful for users

CHANGELOG.md
tstoykov-redis: This comment needs to be removed now

provider/datasource_rediscloud_active_active_subscription_test.go
```diff
@@ -0,0 +1,103 @@
+package provider_test
```
tstoykov-redis: File needs to be moved up into `activeactive` folder

provider/datasource_rediscloud_active_active_subscription_test.go
```diff
@@ -0,0 +1,103 @@
...
+	subscriptionConfig := fmt.Sprintf(utils.GetTestConfig(t, aaSubscriptionDataSourceBoilerplatePath), name)
+	dataSourceConfig := utils.GetTestConfig(t, aaSubscriptionDataSourceConfigPath)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:                 envchecks.ComposePreChecks(t, envchecks.RedisCloudCheck),
+		ProtoV5ProviderFactories: testhelpers.ProtoV5ProviderFactories(),
+		CheckDestroy:             testAccCheckActiveActiveSubscriptionDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: subscriptionConfig,
```
tstoykov-redis: Discussion: Counter suggestion for building the config here, using a mix of ConfigFile and ConfigDirectory instead of inline fetching and formatting of the files. Keeps similar visibility to what we have with our other tests

provider/customtypes/pricing_type.go
```diff
@@ -0,0 +1,175 @@
...
+	_ basetypes.ListTypable  = PricingListType{}
+	_ basetypes.ListValuable = PricingListValue{}
+)
+
+// PricingModel is the typed, tfsdk-tagged view of a pricing entry and the single source of
+// truth for its shape: pricingAttrTypes is derived from the tfsdk taged fields (see below), and it's the
+// compile-checked surface for building each entry (newPricingObject) and for decoding the
+// list (PricingListValue.ElementsAs into []PricingModel).
+// Note that the tfsdk tags must still agree with the pricing block's Attributes in the TF schema.
+type PricingModel struct {
```
tstoykov-redis: After the discussions today, I think `pricing` doesn't need the complication from custom types since it is always computed, both in datasource and resource. Best we do leave it as a simple flatten* function, and only have the `PricingModel` as part of the AA subscription model, so we still get native GO types. But the comment still stands for `maintenance_windows` since that will be reusable in the resource later

tstoykov-redis (changes requested):

provider/datasource_rediscloud_active_active_subscription_test.go
maldimirov: Agree, all new and migrated tests should use the ConfigFile and ConfigValues instead of fmt.Sprintf()

CHANGELOG.md
maldimirov: Removed.

provider/utils/subscription.go
maldimirov: As discussed, the pricing custom types are reverted to generic types. The Model is retained, so that attributes are derived from it and string typed keys are not present in the `flattenPricing()` code.

provider/datasource_rediscloud_active_active_subscription_test.go
maldimirov: Done

provider/datasource_rediscloud_active_active_subscription_test.go
maldimirov: Done.

provider/customtypes/pricing_type.go
maldimirov: ok, done. check it out

provider/activeactive/testdata/active_active_subscription_datasource/datasource.tf
tstoykov-redis: no need for redundant folder naming, can just be `datasource`

provider/activeactive/testdata/active_active_subscription_datasource/subscription.tf
tstoykov-redis: same about redundant folder naming

docs/data-sources/rediscloud_active_active_subscription.md
maldimirov: done

provider/customtypes/attr_types.go
```diff
@@ -7,12 +7,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 )
 
-// attrTypesOf builds an attr-type map from a flat, tfsdk-tagged model struct: each
-// field's `tfsdk` tag becomes the key and the field's own attr.Type the value. A custom
```
tstoykov-redis: I would leave this line about `tfsdk` in, for easier reference in the future

PR #948, opened 2026-07-27
tests: Pin Redis version to 7.4 to use RediSearch module

steliyan:


tstoykov-redis (approved):
change title to reflect use of redis module, which was causing the test to fail

PR #955, opened 2026-07-28
Expand the general provider tests

maldimirov:
Expands the generalized provider test to also test the Framework provider and the "muxed" providers

### DIFF
--- a/provider/provider_test.go
+++ b/provider/provider_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestProvider(t *testing.T) {
-	if err := provider.NewSdkProvider("dev")().InternalValidate(); err != nil {
+	if err := provider.NewSdkProvider("99.99.99")().InternalValidate(); err != nil {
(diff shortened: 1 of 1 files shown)


provider/provider_test.go
tstoykov-redis: let's change the version here from `dev` to `99.99.99` similar to `ProtoV5ProviderFactories`, for consistency

tstoykov-redis (changes requested):

provider/provider_test.go
maldimirov: Updated the Framework and mux provider initialization to use `99.99.99`, consistent with `ProtoV5ProviderFactories`. Thanks for catching this.

maldimirov: Correction: I had initially missed the exact reviewed line. `TestProvider` now also passes `99.99.99` to `NewSdkProvider`, so the requested consistency change is applied there as well.

tstoykov-redis (approved):

PR #960, opened 2026-07-29
tests: Prefix unit tests with "TestUnit"

steliyan:
SDKv2 always has the defined attributes present, so for production code `ram_percentage = 0`.


maldimirov (approved):

provider/pro/resource_rediscloud_pro_subscription.go
maldimirov, line 1: NIT: Maybe change the title of the PR, since there are non-test related changes as well

steliyan: True, but in practice it's test-only as SDKv2 would always provide the `ram_percentage=0`.

PR #984, opened 2026-08-11
tests: Add Active-Active Database unit tests with fully mocked backend (Marin's version)

maldimirov:
Based on https://github.com/RedisLabs/terraform-provider-rediscloud/pull/980
The layering and the behaviour are Steliyan's. This pass over the test harness changes naming, removes some machinery, and fixes a few sharp edges. No test changed its assertions. The same seven tests run under `resource.UnitTest` with plain `go test ./...`, in about eight seconds under `-race`.

## Naming

`MockAPI` becomes `FakeAPI`, and the files, the fixture type and its database type follow. By the usual test-double taxonomy this is a fake, meaning a working in-memory implementation. A mock verifies pre-programmed expectations, which this one does not do. It's also easier than having two things named `Mock`. The methods are renamed on the same principle, to say what they do: `Locked` is now `WithHandlersPaused`, and `NewTask` is now `RegisterTask`.

Doc comments move from floating blocks above the imports onto the types they describe, so that they are closer to the implementations that they document and so that godoc shows them.

## Simplification

- **One mutex instead of two.** The second mutex guarded a request log and the task map. No test used the request log, so it is gone, and the task map is only ever touched under the remaining mutex.
- **No `github.com/google/uuid` dependency.** Task ids come from a counter. Neither the SDK client nor the provider cares what shape they have, so the module graph can stay as it was.
- **One factory helper instead of two.** `FrameworkProviderFactories` takes the provider data as a parameter. The variant that passed nil had no callers.
- **The fixture is ready once constructed.** `newAAAPI` allocates the subscription and registers the handlers, so there is no two-call sequence to get right, and `newAAFixture` returns two values rather than three. `NewAPIClient` takes the `*testing.T` and fails the test itself, so its callers no longer need `require.NoError`.

## Sharp edges

- **A panicking handler used to deadlock the run.** The handler mutex was released without `defer`, so a panic left it locked and every later handler hung. It now unlocks with `defer` and turns the panic into a 500 carrying the stack trace, so the test fails with the panic instead of hanging. `WithHandlersPaused` had the same problem.
- **`RoundTrip` did not close the request body.** The `http.RoundTripper` contract requires it.
- **The `/v1` prefix was trimmed without checking for it.** If the SDK client's base URL ever stops carrying that prefix, no pattern matches and every test fails with "no handler for ...". `RoundTrip` now fails with the actual reason instead.
- **`getRequestedUpgrades` returned the live slice.** It returns a copy now, so a caller cannot watch it change underneath.

## Comments

The long notes are shorter and re-flowed into plain prose. One of them cited line numbers in `resource_active_active_database.go`, and that reference is gone, since it named the wrong lines and would have rotted anyway. The note on the downgrade bug now says outright what the test does and does not establish. It shows that the provider sends the request. It does not show that production state survives one, because that depends on how the real API answers a downgrade target, which this fixture only assumes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


provider/activeactive/fake_aa_api_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+type aaAPI struct {
+	fake *testhelpers.FakeAPI
+
+	nextID int
+	subID  int
+	// database is nil before create and after delete. It has to become nil on delete, because
+	// deleteDatabase waits for the API to answer 404, and a fixture that kept answering 200 would spin until
+	// that wait timed out.
+	database *databaseState
```
tstoykov-redis, line 50: Lets replace this with the actual `ActiveActiveDatabaseModel` and use the actual fields in the GET/PUT/POST

tstoykov-redis (changes requested):

provider/activeactive/fake_aa_api_test.go
maldimirov: Done in [this commit](https://github.com/RedisLabs/terraform-provider-rediscloud/pull/984/changes/0158c1a38a2e7091ea94b6cec53b51b83b29853e). I you don't like it, the whole commit can be discarded/reverted.

PR #1007, opened 2026-08-26
chore: Use user configured timeout, instead of hardcoded one

maldimirov:
The current change is functionally pointless, kind of, since the calling operation timeout will always fire first. I.e. The `Create` or `Update`'s `ctx, cancel := context.WithTimeout(ctx, createTimeout)`. The main goal is not to use the hardcoded SafetyTimeout, so at least that is achieved.

### DIFF
--- a/provider/datasource_rediscloud_active_active_transit_gateway.go
+++ b/provider/datasource_rediscloud_active_active_transit_gateway.go
@@ -153,7 +153,7 @@ func dataSourceActiveActiveTransitGatewayRead(ctx context.Context, d *schema.Res
 		}
 	} else {
 		// No waiting - use existing behaviour that waits for resource to be available
-		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api)
+		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api, utils.TransitGatewayProvisioningTimeout)
 		if err != nil {
 			return diag.FromErr(err)
 		}
--- a/provider/datasource_rediscloud_active_active_transit_gateway.go
+++ b/provider/datasource_rediscloud_active_active_transit_gateway.go
@@ -153,7 +153,7 @@ func dataSourceActiveActiveTransitGatewayRead(ctx context.Context, d *schema.Res
 		}
 	} else {
 		// No waiting - use existing behaviour that waits for resource to be available
-		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api, utils.TransitGatewayProvisioningTimeout)
+		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api, d.Timeout(schema.TimeoutRead))
 		if err != nil {
 			return diag.FromErr(err)
 		}
--- a/provider/privatelink/resource_rediscloud_active_active_private_link.go
+++ b/provider/privatelink/resource_rediscloud_active_active_private_link.go
@@ -182,7 +182,7 @@ func resourceRedisCloudActiveActivePrivateLinkCreate(ctx context.Context, d *sch
 		return diag.FromErr(err)
 	}
 
-	err = waitForActiveActivePrivateLinkToBeActive(ctx, api, subId, regionId)
+	err = waitForActiveActivePrivateLinkToBeActive(ctx, api, subId, regionId, d.Timeout(schema.TimeoutCreate))
 
 	if err != nil {
 		return diag.FromErr(err)
--- a/provider/privatelink/resource_rediscloud_active_active_private_link.go
+++ b/provider/privatelink/resource_rediscloud_active_active_private_link.go
@@ -200,7 +200,7 @@ func resourceRedisCloudActiveActivePrivateLinkCreate(ctx context.Context, d *sch
 	//	return diag.FromErr(err)
 	//}
 
-	err = utils.WaitForSubscriptionToBeActive(ctx, subId, api)
+	err = utils.WaitForSubscriptionToBeActive(ctx, subId, api, d.Timeout(schema.TimeoutCreate))
 	if err != nil {
 		return diag.FromErr(err)
 	}
@@ -360,7 +360,7 @@ func resourceRedisCloudActiveActivePrivateLinkDelete(ctx context.Context, d *sch
 
 	d.SetId("")
 
-	err = utils.WaitForSubscriptionToBeActive(ctx, subId, api)
+	err = utils.WaitForSubscriptionToBeActive(ctx, subId, api, d.Timeout(schema.TimeoutDelete))
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/provider/resource_rediscloud_active_active_private_service_connect.go
+++ b/provider/resource_rediscloud_active_active_private_service_connect.go
@@ -83,7 +83,7 @@ func resourceRedisCloudActiveActivePrivateServiceConnectCreate(ctx context.Conte
 		return diag.FromErr(err)
 	}
 
-	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api)
+	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api, d.Timeout(schema.TimeoutCreate))
 	if err != nil {
 		return diag.FromErr(err)
 	}
@@ -155,7 +155,7 @@ func resourceRedisCloudActiveActivePrivateServiceConnectDelete(ctx context.Conte
 
 	d.SetId("")
 
-	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api)
+	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api, d.Timeout(schema.TimeoutDelete))
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/provider/resource_rediscloud_active_active_private_service_connect.go
+++ b/provider/resource_rediscloud_active_active_private_service_connect.go
@@ -78,7 +78,7 @@ func resourceRedisCloudActiveActivePrivateServiceConnectCreate(ctx context.Conte
 
 	err = waitForPrivateServiceConnectServiceToBeActive(ctx, func() (result interface{}, state string, err error) {
 		return refreshPrivateServiceConnectServiceActiveActiveStatus(ctx, subscriptionId, regionId, api)
-	})
+	}, d.Timeout(schema.TimeoutCreate))
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/provider/resource_rediscloud_active_active_private_service_connect.go
+++ b/provider/resource_rediscloud_active_active_private_service_connect.go
@@ -59,6 +59,7 @@ func resourceRedisCloudActiveActivePrivateServiceConnect() *schema.Resource {
 
 func resourceRedisCloudActiveActivePrivateServiceConnectCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	api := meta.(*client.ApiClient)
+	timeout := d.Timeout(schema.TimeoutCreate)
 
 	subscriptionId, err := strconv.Atoi(d.Get("subscription_id").(string))
 	if err != nil {
@@ -78,12 +79,12 @@ func resourceRedisCloudActiveActivePrivateServiceConnectCreate(ctx context.Conte
 
 	err = waitForPrivateServiceConnectServiceToBeActive(ctx, func() (result interface{}, state string, err error) {
 		return refreshPrivateServiceConnectServiceActiveActiveStatus(ctx, subscriptionId, regionId, api)
-	}, d.Timeout(schema.TimeoutCreate))
+	}, timeout)
 	if err != nil {
 		return diag.FromErr(err)
 	}
 
-	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api, d.Timeout(schema.TimeoutCreate))
+	err = utils.WaitForSubscriptionToBeActive(ctx, subscriptionId, api, timeout)
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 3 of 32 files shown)


provider/privatelink/resource_rediscloud_active_active_private_link.go
tstoykov-redis: Not for this PR, but we need to review the update logic, as it probably needs a wait* block added, since the API returns a task for that as well

provider/resource_rediscloud_active_active_private_service_connect_endpoint.go
tstoykov-redis: This also needs handling

provider/utils/get_set.go
```diff
@@ -7,14 +7,8 @@ import (
...
-// Subscriptions and Databases. Reads and Deletions have their own, stricter timeouts because they consistently behave
-// well. The Terraform operation-level timeout should kick in way before we hit this and kill the task.
-// Unfortunately there's no "time-remaining-before-timeout" utility, or we could use that in the wait blocks.
-const SafetyTimeout = 6 * time.Hour
-
-// TransitGatewayProvisioningTimeout is used when waiting for Transit Gateway resources to become available during
-// subscription provisioning. This is shorter than SafetyTimeout as tests typically complete within 45 minutes.
+// TransitGatewayProvisioningTimeout limits waits for Transit Gateway resources during subscription provisioning.
+// The limit keeps these waits within the 45-minute duration in which acceptance tests typically finish.
 const TransitGatewayProvisioningTimeout = 40 * time.Minute
```
tstoykov-redis: This can also be removed

provider/datasource_rediscloud_active_active_transit_gateway.go
```diff
@@ -153,7 +153,7 @@ func dataSourceActiveActiveTransitGatewayRead(ctx context.Context, d *schema.Res
 		}
 	} else {
 		// No waiting - use existing behaviour that waits for resource to be available
-		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api)
+		tgwTask, err := utils.WaitForActiveActiveTransitGatewayResourceToBeAvailable(ctx, subId, regionId, api, utils.TransitGatewayProvisioningTimeout)
```
tstoykov-redis: we can pass the read timeout directly, since anything exceeding it, will timeout, and otherwise, we once again have a static timeout

provider/resource_rediscloud_active_active_private_service_connect.go
```diff
@@ -78,12 +78,12 @@ func resourceRedisCloudActiveActivePrivateServiceConnectCreate(ctx context.Conte
 
 	err = waitForPrivateServiceConnectServiceToBeActive(ctx, func() (result interface{}, state string, err error) {
 		return refreshPrivateServiceConnectServiceActiveActiveStatus(ctx, subscriptionId, regionId, api)
-	})
+	}, d.Timeout(schema.TimeoutCreate))
```
tstoykov-redis: Lets extract this to a variable for code consistency

provider/resource_rediscloud_active_active_private_service_connect_endpoint_accepter.go
```diff
@@ -70,6 +70,16 @@ func resourceRedisCloudActiveActivePrivateServiceConnectEndpointAccepter() *sche
 }
 
 func resourceRedisCloudActiveActivePrivateServiceConnectEndpointAccepterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	timeout := d.Timeout(schema.TimeoutCreate)
+	return resourceRedisCloudActiveActivePrivateServiceConnectEndpointAccepterApply(ctx, d, meta, timeout)
```
tstoykov-redis: we can just pass the `d.Timeout` directly here

provider/resource_rediscloud_active_active_subscription_regions.go
```diff
@@ -137,10 +137,16 @@ func resourceRedisCloudActiveActiveRegionCreate(ctx context.Context, d *schema.R
 	}
 	d.SetId(strconv.Itoa(subId))
 
-	return resourceRedisCloudActiveActiveRegionUpdate(ctx, d, meta)
+	timeout := d.Timeout(schema.TimeoutCreate)
```
tstoykov-redis: Again, can directly pass this here, no need to create the variable for a single use

provider/resource_rediscloud_active_active_transit_gateway_attachment.go
```diff
@@ -116,17 +117,22 @@ func resourceRedisCloudActiveActiveTransitGatewayAttachmentCreate(ctx context.Co
...
 	}
 
 	d.SetId(transitgateway.BuildActiveActiveTransitGatewayAttachmentId(subscriptionId, regionId, tgwId))
 
-	return resourceRedisCloudActiveActiveTransitGatewayAttachmentRead(ctx, d, meta)
+	return resourceRedisCloudActiveActiveTransitGatewayAttachmentReadWithTimeout(ctx, d, meta, timeout)
 }
 
 func resourceRedisCloudActiveActiveTransitGatewayAttachmentRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	timeout := d.Timeout(schema.TimeoutRead)
```
tstoykov-redis: And here, we can just directly pass the `d.Timeout`

provider/resource_rediscloud_active_active_transit_gateway_route.go
```diff
@@ -88,10 +89,15 @@ func resourceRedisCloudActiveActiveTransitGatewayRouteCreate(ctx context.Context
 
 	d.SetId(transitgateway.BuildActiveActiveTransitGatewayAttachmentId(subId, regionId, tgwId))
 
-	return resourceRedisCloudActiveActiveTransitGatewayRouteRead(ctx, d, meta)
+	return resourceRedisCloudActiveActiveTransitGatewayRouteReadWithTimeout(ctx, d, meta, timeout)
 }
 
 func resourceRedisCloudActiveActiveTransitGatewayRouteRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	timeout := d.Timeout(schema.TimeoutRead)
```
tstoykov-redis: And here, pass directly

provider/resource_rediscloud_private_service_connect.go
```diff
@@ -71,12 +71,12 @@ func resourceRedisCloudPrivateServiceConnectCreate(ctx context.Context, d *schem
 
 	err = waitForPrivateServiceConnectServiceToBeActive(ctx, func() (result interface{}, state string, err error) {
 		return refreshPrivateServiceConnectServiceStatus(ctx, subscriptionId, api)
-	})
+	}, d.Timeout(schema.TimeoutCreate))
```
tstoykov-redis: Here we can extract it, since it's used multiple times in the method

provider/resource_rediscloud_private_service_connect_endpoint_accepter.go
```diff
@@ -65,6 +65,16 @@ func resourceRedisCloudPrivateServiceConnectEndpointAccepter() *schema.Resource
 }
 
 func resourceRedisCloudPrivateServiceConnectEndpointAccepterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	timeout := d.Timeout(schema.TimeoutCreate)
+	return resourceRedisCloudPrivateServiceConnectEndpointAccepterApply(ctx, d, meta, timeout)
```
tstoykov-redis: Here we can directly pass the timeout again

tstoykov-redis (changes requested):

provider/utils/get_set.go
maldimirov: done

provider/resource_rediscloud_active_active_private_service_connect.go
maldimirov: done

provider/resource_rediscloud_active_active_private_service_connect_endpoint_accepter.go
maldimirov: done

provider/datasource_rediscloud_active_active_transit_gateway.go
maldimirov: done

provider/resource_rediscloud_active_active_subscription_regions.go
maldimirov: done

provider/resource_rediscloud_active_active_transit_gateway_attachment.go
maldimirov: done

provider/resource_rediscloud_active_active_transit_gateway_route.go
maldimirov: done

provider/resource_rediscloud_private_service_connect.go
maldimirov: done

provider/resource_rediscloud_private_service_connect_endpoint_accepter.go
maldimirov: done

provider/datasource_rediscloud_active_active_transit_gateway.go
tstoykov-redis, line 115: Just saw this now, but this big if/else, can be replaced now, since the actual retry logic is the same, but the old method didn't have the signature to pass a timeout

tstoykov-redis: Also if possible, refactor the one in `datasource_rediscloud_transit_gateway.go` as it has the same logic, and we can extract that method into a proper wait function

tstoykov-redis (changes requested):

provider/datasource_rediscloud_active_active_transit_gateway.go
```diff
@@ -45,11 +47,12 @@ func dataSourceActiveActiveTransitGateway() *schema.Resource {
...
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     0,
+				Deprecated: "Configure the `read` timeout in the `timeouts` block instead. " +
+					"This attribute will be removed in version 3.0 of the provider.",
```
tstoykov-redis: Lets use ` This attribute will be removed in the next major version of the provider.` instead, to keep consistent wording across the project

provider/datasource_rediscloud_active_active_transit_gateway.go
```diff
@@ -45,11 +47,12 @@ func dataSourceActiveActiveTransitGateway() *schema.Resource {
 				Computed:    true,
 			},
 			"wait_for_tgw_timeout": {
-				Description: "When set, retry fetching until a Transit Gateway matching the filters is found or the specified timeout (in seconds) is reached. " +
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
```
tstoykov-redis: Just add a note, that it will not exceed the max timeout for the Read operation, since we are changing it. Also update the matching .md file

provider/datasource_rediscloud_transit_gateway.go
```diff
@@ -40,11 +42,12 @@ func dataSourceTransitGateway() *schema.Resource {
 				Computed:    true,
 			},
 			"wait_for_tgw_timeout": {
-				Description: "When set, retry fetching until a Transit Gateway matching the filters is found or the specified timeout (in seconds) is reached. " +
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
```
tstoykov-redis: Same comments here about .md file and deprecated message

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+			}
+			if requestErr != nil {
+				return nil, "", requestErr
+			}
+
+			if tgwTask == nil || tgwTask.Response == nil || tgwTask.Response.Resource == nil {
+				return nil, waitStateWaiting, nil
+			}
+
+			if !transitGatewayTaskMatchesFilters(tgwTask, filters) {
```
tstoykov-redis: we can merge this condition into the one above

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+				tgwTask, requestErr = api.Client.TransitGatewayAttachments.GetActiveActive(ctx, subId, regionId)
+			} else {
+				tgwTask, requestErr = api.Client.TransitGatewayAttachments.Get(ctx, subId)
+			}
+			if requestErr != nil {
+				return nil, "", requestErr
+			}
+
+			if tgwTask == nil || tgwTask.Response == nil || tgwTask.Response.Resource == nil {
+				return nil, waitStateWaiting, nil
```
tstoykov-redis: we should return an empty value for the result, not nil, to avoid the exhausting the `NotFoundChecks`, and erroring out, before we time out. Something like `struct{}{}` should do the job

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+) (*attachments.GetAttachmentsTask, error) {
+	gatewayName := "Transit Gateway"
+	location := fmt.Sprintf("subscription %d", subId)
+	if subscriptionType == subscriptions.SubscriptionDeploymentTypeActiveActive {
+		gatewayName = "Active-Active Transit Gateway"
+		location = fmt.Sprintf("subscription %d, region %d", subId, regionId)
+	}
+
+	wait := &retry.StateChangeConf{
+		Pending:      []string{waitStateWaiting},
```
tstoykov-redis: As both the Pending and Target statuses are randomly selected by us, and do not reflect the actual values returned from the API, we should be able to reuse this method instead of `WaitForTransitGatewayAttachmentToBeAvailable` and `WaitForActiveActiveTransitGatewayResourceToBeAvailable` to clean up even more code

tstoykov-redis (changes requested):

provider/datasource_rediscloud_active_active_transit_gateway.go
tstoykov-redis: All of these are for a separate PR as agreed

provider/datasource_rediscloud_active_active_transit_gateway.go
tstoykov-redis: As agreed, separate PR

tstoykov-redis (approved):

provider/datasource_rediscloud_active_active_transit_gateway.go
maldimirov: Yep, here it is https://github.com/RedisLabs/terraform-provider-rediscloud/pull/1020

PR #1018, opened 2026-09-14
Fix PrivateLink principals comparison by value

### LINKED ISSUE #1013: `rediscloud_private_link`: every principal update fails with 409 `PRIVATE_LINK_PRINCIPAL_ALREADY_EXISTS` — principal diff compares pointers, not values
**Provider version:** v2.18.1; still present in **v2.19.0** (verified by reading the source at tag `v2.19.0`, not by running it)
**Resource:** `rediscloud_private_link`
**Terraform version:** 1.15.x

## Summary

Once a `rediscloud_private_link` has at least one principal, **any** `terraform apply` that changes the `principal` set fails with:

```text
409 CONFLICT - PRIVATE_LINK_PRINCIPAL_ALREADY_EXISTS: Principal already exists in private link configuration
```

This is not limited to additions. Removing a principal fails too: going from `[A, B]` to `[B]` still puts B in the create list, B already exists, and the apply aborts. In practice the principal set of an existing PrivateLink cannot be changed through Terraform at all.

This affects the ordinary, fully documented `aws_account` principal path.

## Root cause

`findPrincipalsToCreate` and `findPrincipalsToDelete` in [`provider/privatelink/resource_rediscloud_private_link.go`](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/v2.19.0/provider/privatelink/resource_rediscloud_private_link.go) compare principals with:

```go
if tfPrincipal.Principal == apiPrincipal.Principal {   // findPrincipalsToCreate
if apiPrincipal.Principal == tfPrincipal.Principal {   // findPrincipalsToDelete
```

Both fields are `*string`, so this compares **pointer identity, not string values**. The two sides are built from different allocations — the Terraform side by `principalsFromSet` in [`provider/privatelink/utils.go`](https://github.com/RedisLabs/terraform-provider-rediscloud/blob/v2.19.0/provider/privatelink/utils.go):

```go
Principal: redis.String(principalMap["principal"].(string)),
```

and the API side by unmarshalling the API response. The pointers are therefore never equal
and both comparisons are always `false`. Consequently:

- **`findPrincipalsToCreate` returns every configured principal**, including ones already live. The first `POST …/private-link/principals` for an existing principal returns 409 and the apply aborts.
- **`findPrincipalsToDelete` returns every pre-existing principal.** Creates run before deletes in `resourceRedisCloudPrivateLinkUpdate` and the create failure aborts the update,  so this path is currently masked. If duplicate creates were ever made idempotent — or the  create phase otherwise did not abort — the provider would delete principals the user never removed from configuration, revoking and re-issuing their RAM invitations and disassociating live consumer endpoints as a side effect of an unrelated addition.

## Reproduction

Apply a PrivateLink with a single principal:

```hcl
resource "rediscloud_private_link" "example" {
  subscription_id = rediscloud_subscription.example.id
  share_name      = "redis-sub-000000"

  principal {
    principal       = "111111111111"
    principal_type  = "aws_account"
    principal_alias = "account A"
  }
}
```

This succeeds. Now add a second principal:

```hcl
  principal {
    principal       = "222222222222"
    principal_type  = "aws_account"
    principal_alias = "account B"
  }
```

`terraform apply` fails:

```text
Error: failed when creating PrivateLink 0: 409 CONFLICT - PRIVATE_LINK_PRINCIPAL_ALREADY_EXISTS: Principal already exists in private link configuration
```

Index 0 is principal A: `principalsFromSet` ends with

```go
sort.Slice(createPrincipals, func(i, j int) bool {
    return *createPrincipals[i].Principal < *createPrincipals[j].Principal
})
```

so the create list is ordered by principal string and the provider re-POSTs the already-existing A before ever reaching B.

Because of that ordering, a genuinely new principal can be created *before* the apply fails on an existing one — so **the update can be left partially applied**, with the successful creates committed server-side and not recorded in Terraform state.

## Expected behaviour

```text
API state: A     config: A, B    → create: [B]   delete: []
API state: A, B  config: B       → create: []    delete: [A]
API state: A, B  config: A, B    → create: []    delete: []
```

Unchanged principals are neither POSTed nor deleted because some other principal changed.

## Suggested fix

Dereference before comparing, nil-safely, in both functions:

```go
if redis.StringValue(tfPrincipal.Principal) == redis.StringValue(apiPrincipal.Principal) {
```

`redis.StringValue` is already used elsewhere in the provider. With that change, unchanged principals fall out of both the create and the delete list, which also resolves the latent delete-and-recreate behaviour above.

Regression tests worth having, over the transitions `[] → [A]`, `[A] → [A, B]`,
`[A, B] → [B]`, `[A, B] → [A, B]`, `[A, B] → [A, B, C]`. The load-bearing assertion for the
unchanged case is that A is neither POSTed nor deleted when only another principal changes.

## Workaround

For anyone hitting this before the fix ships: create the new principal **out of band via the API** with the exact alias the configuration will use, verify it actually appears in the share's `principals[]`, and only then merge the matching `principal` block. The refreshed plan then carries no principal diff, so the broken update path is never entered. Never let an apply carry a principal diff against a non-empty live principal set.

maldimirov:
## Summary

- compare PrivateLink principal values rather than pointer identities when calculating additions and removals
- add unit coverage for retaining, adding, and removing principals

## Testing

- go test ./provider/privatelink -run ^TestPrincipalChanges$ -count=1
- go test ./provider/privatelink -count=1

Fixes #1013


provider/privatelink/principal_diff_test.go
```diff
@@ -0,0 +1,89 @@
...
+			assert.Equal(t, test.principalsToAdd, findPrincipalsToCreate(test.apiPrincipals, test.tfPrincipals),
+				"creation diff must contain exactly the principals present in Terraform configuration but absent from API state")
+			assert.Equal(t, test.principalsToDrop, findPrincipalsToDelete(test.apiPrincipals, test.tfPrincipals),
+				"deletion diff must contain exactly the principals present in API state but absent from Terraform configuration")
+		})
+	}
+}
+
+func apiPrincipalList(principals ...string) []*pl.PrivateLinkPrincipal {
+	result := make([]*pl.PrivateLinkPrincipal, 0, len(principals))
```
tstoykov-redis, line 55: Why not reuse the `terraformPrincipalList` function, and just iterate and `&value` the list generated from it, instead of duplicating code

tstoykov-redis (changes requested):

provider/privatelink/principal_diff_test.go
maldimirov: Done.

PR #1020, opened 2026-09-16
chore: Unify Transit Gateway data source wait behavior

maldimirov:
Stacks on #1007.

## Summary
- make both Pro and Active-Active Transit Gateway data sources poll until their filters match
- use configurable read timeouts, with deprecated `wait_for_tgw_timeout` as a compatibility override until provider v3.0
- share the polling logic while selecting the SDK endpoint from the subscription deployment type

## Testing
- `GOCACHE=/tmp/terraform-provider-rediscloud-go-build go test ./provider/...`


provider/datasource_rediscloud_active_active_transit_gateway.go
```diff
@@ -45,11 +47,12 @@ func dataSourceActiveActiveTransitGateway() *schema.Resource {
...
-				Description: "When set, retry fetching until a Transit Gateway matching the filters is found or the specified timeout (in seconds) is reached. " +
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     0,
+				Deprecated: "Configure the `read` timeout in the `timeouts` block instead. " +
```
tstoykov-redis, line 54: Lets use  `This attribute will be removed in the next major version of the provider.` instead, to keep consistent wording across the project

provider/datasource_rediscloud_active_active_transit_gateway.go
```diff
@@ -45,11 +47,12 @@ func dataSourceActiveActiveTransitGateway() *schema.Resource {
 				Computed:    true,
 			},
 			"wait_for_tgw_timeout": {
-				Description: "When set, retry fetching until a Transit Gateway matching the filters is found or the specified timeout (in seconds) is reached. " +
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
```
tstoykov-redis, line 50: Just add a note, that it will not exceed the max timeout for the Read operation, since we are changing it. Also update the matching .md file

provider/datasource_rediscloud_transit_gateway.go
```diff
@@ -40,11 +42,12 @@ func dataSourceTransitGateway() *schema.Resource {
 				Computed:    true,
 			},
 			"wait_for_tgw_timeout": {
-				Description: "When set, retry fetching until a Transit Gateway matching the filters is found or the specified timeout (in seconds) is reached. " +
-					"Useful when accepting a TGW invitation and querying the TGW in the same Terraform run.",
-				Type:     schema.TypeInt,
-				Optional: true,
-				Default:  0,
+				Description: "Overrides the read timeout, in seconds, while waiting for a Transit Gateway matching the configured filters.",
```
tstoykov-redis, line 45: Same comments here about .md file and deprecated message

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+			}
+			if requestErr != nil {
+				return nil, "", requestErr
+			}
+
+			if tgwTask == nil || tgwTask.Response == nil || tgwTask.Response.Resource == nil {
+				return nil, waitStateWaiting, nil
+			}
+
+			if !transitGatewayTaskMatchesFilters(tgwTask, filters) {
```
tstoykov-redis, line 167: we can merge this condition into the one above

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+				tgwTask, requestErr = api.Client.TransitGatewayAttachments.GetActiveActive(ctx, subId, regionId)
+			} else {
+				tgwTask, requestErr = api.Client.TransitGatewayAttachments.Get(ctx, subId)
+			}
+			if requestErr != nil {
+				return nil, "", requestErr
+			}
+
+			if tgwTask == nil || tgwTask.Response == nil || tgwTask.Response.Resource == nil {
+				return nil, waitStateWaiting, nil
```
tstoykov-redis, line 164: we should return an empty value for the result, not nil, to avoid the exhausting the `NotFoundChecks`, and erroring out, before we time out. Something like `struct{}{}` should do the job

provider/utils/wait.go
```diff
@@ -115,6 +115,78 @@ func WaitForDatabaseToBeActive(ctx context.Context, subId, id int, api *client.A
...
+) (*attachments.GetAttachmentsTask, error) {
+	gatewayName := "Transit Gateway"
+	location := fmt.Sprintf("subscription %d", subId)
+	if subscriptionType == subscriptions.SubscriptionDeploymentTypeActiveActive {
+		gatewayName = "Active-Active Transit Gateway"
+		location = fmt.Sprintf("subscription %d, region %d", subId, regionId)
+	}
+
+	wait := &retry.StateChangeConf{
+		Pending:      []string{waitStateWaiting},
```
tstoykov-redis, line 143: As both the Pending and Target statuses are randomly selected by us, and do not reflect the actual values returned from the API, we should be able to reuse this method instead of `WaitForTransitGatewayAttachmentToBeAvailable` and `WaitForActiveActiveTransitGatewayResourceToBeAvailable` to clean up even more code

tstoykov-redis (changes requested):

